PR #13, opened 2019-05-13
Reference to STM32L432 driver implementation

nickray:
FYI: Finally got my port to STM32L432 to run the `serial` example. Cross-referencing https://github.com/dmitrystu/libusb_stm32 was quite helpful. Not sure at all about the naming, as L432, L442, L433, L443 all seem to be the same (apart from LCD/AES additions).

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ Related crates
 
 * [stm32f103xx-usb](https://github.com/mvirkkunen/stm32f103xx-usb) - device-driver implementation
   for STM32F103 microcontrollers. Also contains runnable examples.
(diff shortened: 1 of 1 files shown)


mvirkkunen:
Very nice!

What comes to naming, in the future it might even make sense to try to merge all the drivers for this USB peripheral into one crate. libopencm3 seems to have managed to share a lot of the code between the drivers. I called my crate "stm32f103xx-usb" because that's the only board with the peripheral I have at the moment, but the same crate would likely work with other F1 chips.

Having an implementation for a different version of the peripheral is very good because it's a starting point to diff and find out how different it is - and see if it's feasible to try to merge the different drivers or if it would just result in an unreadable amount of `#[cfg(feature)]`'s.

PR #16, opened 2019-07-22
[breaking change] Add new endpoint status methods

Disasm:


Disasm:
Implementation: https://github.com/stm32-rs/stm32-usbd/tree/ep-api

src/bus.rs
```diff
@@ -130,6 +130,43 @@ pub trait UsbBus: Sync + Sized {
     }
 }
 
+// TODO: merge this into UsbBus with the next breaking change
```
mvirkkunen: I think it's perfectly fine to just add everything to `UsbBus`, it's versioned, and that's what 0.x versions are for.

src/endpoint.rs
```diff
@@ -107,6 +107,18 @@ impl<B: UsbBus, D: EndpointDirection> Endpoint<'_, B, D> {
     }
 }
 
+impl<B: UsbBusEndpointStatus> Endpoint<'_, B, In> {
+    /// Returns true if the endpoint is ready to write another packet, otherwise returns false.
+    pub fn can_write(&self) -> bool {
+        self.bus().can_write(self.address).expect("unexpected error")
```
mvirkkunen: I think it would be better to return errors and not panic. As you can see the built-in endpoint objects are kind of becoming dumb wrappers for `UsbBus` which is why I'm considering getting rid of them.

src/endpoint.rs
```diff
@@ -107,6 +107,18 @@ impl<B: UsbBus, D: EndpointDirection> Endpoint<'_, B, D> {
...
+    }
+
+    /// Returns true if all the endpoint write transfers were finished, otherwise returns false.
+    pub fn write_complete(&self) -> bool {
+        self.bus().write_complete(self.address).expect("unexpected error")
```
mvirkkunen: Would prefer to avoid panics here too

src/endpoint.rs
```diff
@@ -128,6 +140,20 @@ impl<B: UsbBus> Endpoint<'_, B, In> {
...
+impl<B: UsbBusEndpointStatus> Endpoint<'_, B, Out> {
+    /// Returns the length of the packet available to read, if any.
+    ///
+    /// Note that this function returns only the length of the first packet, not the total length of
+    /// all the pending packets.
+    pub fn available_read(&self) -> Option<usize> {
+        match self.bus().available_read(self.address) {
+            Ok(size) => Some(size),
+            Err(crate::UsbError::WouldBlock) => None,
+            Err(e) => panic!("unexpected error: {:?}", e),
```
mvirkkunen: And here

mvirkkunen:
The `endpoint-trait` branch will implement "flush" which is the most important - if the other methods are still needed let's think about this again once the branch is merged.

PR #20, opened 2019-07-28
[breaking change] Remove force_reset() in favor of a device-dependent method

Disasm:


Disasm:
This helps to replace BluePill USB reset hack (which doesn't work at the moment anyway) with something more meaningful. This disconnect functionality is supported by newer USB FS peripherals as well as by OTG FS peripherals.
Other option is to remove this functionality completely.

mvirkkunen:
This has been removed in the `endpoint-trait` branch.

PR #29, opened 2020-01-30
Change USB version to 2.1 and add BOS functionality

redpfire:
Hi! I've created an implementation for Binary Object Store(BOS) of USB 2.1 standard.

My main motivation for implementing BOS was the ability to implement and use WebUSB and Microsoft OS capability descriptor (guaranteed detection of the device when using Windows 8.1+). 
I'm working on a crate that supports WebUSB (prototype now working) and I'll be grateful if you merge.
I also followed the documentation style so it all looks uniform. I'm open to any changes.

image


redpfire:
I don't know what happened with Travis.. It did pass after every commit/push but now it hang.

mvirkkunen:
Hi and thanks for the PR! Seems like a good addition to me.

Some questions/improvements that I can think of:

1. Does changing `bcdUSB` affect operating system compatibility? I'd imagine it doesn't but you never know...
2. I would prefer to have a separate "writer" for this. `DescriptorWriter` has a bunch of methods that don't make sense in the context of `get_bos_blobs` and vice-versa, and only allow you to write broken descriptors.
3. `bos_capability` should check that the data fits in the buffer before attempting to write to it and return an error code instead of panicking.
4. `bos_start`/`end_bos` is a bit inconsistent naming-wise :slightly_smiling_face: 

Eventually I still want to do compile-time descriptor generation which would also involve this, and would probably allow for one to create extensions that generate stuff like all the WebUSB related descriptors out of a little bit of input data (think just writing `WebUSB(&["my.website"])` to enable it). But doing this at runtime as it currently is would probably be pretty bloated, so I wouldn't go that far yet.

Once the things above are resolved I'd be happy to merge.

redpfire:
So:

1. It does introduce BOS to the pc. It lets the driver know that it can pull BOS from the device.
2. I'd agree, but BOS is a descriptor after all. And i exposed an interface to create whichever descriptor you'd like under BOS. That's when i introduce `blobs` - a binary data that represents the content of a descriptor in the BOS. They're mostly proprietary - that's why i chose not to make specific templates, so you can choose.
3. I was thinking i implemented that, but it's worth mentioning. I'll work on that for sure!
4. I know, but i chose the naming scheme _you_ chose before me. I chose the naming convention you represented in the code.

redpfire:
Take a look at these two functions right here:
https://github.com/mvirkkunen/usb-device/blob/a44c4ce16fe5a3bfed8b1bfd51e536018f753cdd/src/descriptor.rs#L153

redpfire:
As to the WebUSB descriptors - you actually have to create a class that reacts to specifig `VENDOR` operation to deliver for example the URL of the landing page. As such:
```
impl<B: UsbBus> UsbClass<B> for WebUsb<B> {
    fn get_bos_blobs(&self, w: &mut DescriptorWriter) -> Result<()> {
        w.bos_capability(0x05, &[
            0x38, 0xb6, 0x08, 0x34, 0xa9,
            0x09, 0xa0, 0x47, 0x8b, 0xfd,
            0xa0, 0x76, 0x88, 0x15, 0xb6,
            0x65,  // PlatformCapabilityUUID

            0x00, 0x01, // bcdVersion of WebUSB
            self.bVendorCode, // bVendorCode
            self.iLandingPage, // iLandingPage
        ])
    }

    fn control_in(&mut self, xfer: ControlIn<B>) {
        let req = *xfer.request();

        if !(req.request_type == control::RequestType::Vendor
             && req.recipient == control::Recipient::Device
             && req.request == self.bVendorCode
             && req.value == self.iLandingPage as u16
             )
        {
            return;
        }

        match req.index {
            GET_URL => {
                xfer.accept_with(&self.desc[..self.desc_len]).ok()
            },
            _ => xfer.reject().ok(),
        };
    }
}
```
That's an actual working example right there.

redpfire:
I'd also suggest changing `buf: unsafe { mem::uninitialized() }` into ` buf: unsafe { mem::zeroed() }` to suppress compiler warnings ;)

mvirkkunen:
>1. It does introduce BOS to the pc. It lets the driver know that it can pull BOS from the device.

Yes, however I meant to say to *backwards* compatibility. Do you know if older systems (without USB 3.0 support, for instance) might reject the device because the USB version is reported as "too new"?

>2. I'd agree, but BOS is a descriptor after all. And i exposed an interface to create whichever descriptor you'd like under BOS. That's when i introduce blobs - a binary data that represents the content of a descriptor in the BOS. They're mostly proprietary - that's why i chose not to make specific templates, so you can choose.

The `DescriptorWriter` is only exposed (outside of the crate) for writing configuration descriptors currently so it should have been named `ConfigurationDescriptorWriter` but I was lazy and put other (private) descriptors in there as well. It's not a very good interface to begin with, I really wanted to do something like typestates to make writing malformed descriptors impossible but couldn't get it to work back then because I was a Rust newbie. In the case of `get_bos_blobs` methods such as `interface` or `endpoint` can't possibly be used to write a valid descriptor so it feels to me that it would be weird to have those methods available.

The rest of the changes look good to me!

redpfire:
> Yes, however I meant to say to backwards compatibility. Do you know if older systems (without USB 3.0 support, for instance) might reject the device because the USB version is reported as "too new"?
USB 2.1 is fully backwards compatible to USB 2.0, and has every feature that USB 2.0 has - even it's compatibility. In terms of drivers rejecting the device because it is "too now" - i believe that's impossible. USB 2.1 only signals the driver that it's compatible with the BOS part of USB 3.0 specs.
> In the case of get_bos_blobs methods such as interface or endpoint can't possibly be used to write a valid descriptor so it feels to me that it would be weird to have those methods available.
What name would you suggest for such a writer? `BOSWriter`? I can implement such a writer right now.

mvirkkunen:
`BOSWriter` sounds good to me! It could even go in the same file. The descriptor generator needs to be improved later anyways.

redpfire:
I also have an idea on improving the `bos_capability` right away - unifying UUID implementing it straight into the function. Do you have an idea how that could be improved upon?

mvirkkunen:
What do you mean by "unifying the UUID"? Something like adding a separate method for writing the "Platform descriptors" which contain a UUID (which are probably what most people will want to write - a lot of the others are related to USB internals which should be handled by usb-device itself)

redpfire:
Each capability descriptor has a discrete UUID - `PlatformCapabilityUUID`. For ex. here:
```
fn get_bos_blobs(&self, w: &mut DescriptorWriter) -> Result<()> {
        w.bos_capability(0x05, &[
            0x38, 0xb6, 0x08, 0x34, 0xa9,
            0x09, 0xa0, 0x47, 0x8b, 0xfd,
            0xa0, 0x76, 0x88, 0x15, 0xb6,
            0x65,  // PlatformCapabilityUUID

            0x00, 0x01, // bcdVersion of WebUSB
            self.bVendorCode, // bVendorCode
            self.iLandingPage, // iLandingPage
        ])
    }
```

redpfire:
I also have a problem implementing the `BOSWriter`. I want to give it a `DescriptorWriter` to avoid doubling the code, but the compiler screams at me about the lifetimes (i still havent' figured them out).
```
impl BOSWriter<'_> {
    pub(crate) fn new(writer: &mut DescriptorWriter<'_>) -> BOSWriter<'_> {
        BOSWriter {
            writer: writer,
            num_bos_blobs_mark: None,
        }
    }

    pub(crate) fn bos(&mut self) -> Result<()> {
        self.num_bos_blobs_mark = Some(self.writer.position + 4);
        self.writer.write(
            descriptor_type::BOS,
            &[
                0x00, 0x00, // totalLen
                0x00, // blobsNum
            ])
    }
}
```

mvirkkunen:
Only platform capability descriptor (bDevCapabilitiyType=0x05) seem to have a UUID, not all capability descriptors.

If you need help with lifetimes and such it would be easier if you could join the chat room at https://matrix.to/#/#rust-embedded:matrix.org (or #rust-embedded on Freenode IRC). References to references require some lifetime magic to work.

(Also I just realized - the name should probably be `BosWriter` to stay consistent with capitalization)

redpfire:
> Only platform capability descriptor (bDevCapabilitiyType=0x05) seem to have a UUID, not all capability descriptors.

Yes, but i was thinking about shorting the path and providing the uuid directly in the function call.

> If you need help with lifetimes and such it would be easier if you could join the chat room at https://matrix.to/#/#rust-embedded:matrix.org (or #rust-embedded on Freenode IRC). References to references require some lifetime magic to work.

Thanks! I managed to sort it out though :)

mvirkkunen:
>Yes, but i was thinking about shorting the path and providing the uuid directly in the function call.

So it would only support PLATFORM type capability descriptors that take the UUID and none of the other types?

redpfire:
Oh yeah, i forgot about that! Nevermind then!

mvirkkunen:
Since the PLATFORM type descriptors are probably the most common ones there could be a separate method for writing those that calls the other method inside! Just to make it a little easier.

redpfire:
That's a great idea, yet still i don't exactly know how to implement the UUID in a function!

redpfire:
merged

mvirkkunen:
Merged in https://github.com/mvirkkunen/usb-device/commit/3eb79b62706993688b88b3399a96efcf9b6d07bc but I failed at rebasing

PR #37, opened 2020-03-27
Support high speed operation for test class with feature

conorpp:
We're using usb-device with our peripheral in high speed mode, and needed to increase the buffer sizes in the test class.  

In high speed mode, the max size for bulk packets is 512 bytes, and 1024 for interrupt packets.

Thanks for your work on this awesome Rust USB stack!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,8 @@ rand = "0.6.1"
 [features]
 # Use a 256 byte buffer for control transfers instead of 128.
 control-buffer-256 = []
+# Use larger endpoint buffers for highspeed operation (default fullspeed)
+test-class-high-speed = []
 
 [[test]]
 name = "test_class_host"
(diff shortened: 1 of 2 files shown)


mvirkkunen:
Whoa, how much hacking did you have to do descriptor wise to get high speed working? That's something on my non-existent roadmap but I haven't looked at the details yet.

Would you mind to rename/if I rename the feature to something that makes it clear it's only about TestClass and doesn't actually add high-speed support otherwise? Something like `test-class-high-speed`.

conorpp:
Changed the feature name to `test-class-high-speed`.

Surprisingly, I didn't have to do much to the descriptor to get high speed working.  It was just making the endpoint sizes larger.  With the `test-class-high-speed` flag set, I can run the usb tests just fine on both high speed and full speed interfaces.  To force a full speed interface, I'm using an ancient usb hub that only supports full speed (hard to find nowadays).  Probably with the sizes not changing when fullspeed is used, this could cause problems for some applications, not sure.

mvirkkunen:
There's probably something funky going on in full speed mode, such as the operating system ignoring the invalid descriptors and assuming some other packet size. I'd imagine this would at least cause some warnings when enumerating, but if this is helpful for for testing (since real support isn't there yet), I suppose it's OK to merge this with an explanation as to what it's for.

PR #40, opened 2020-05-31
Fix tests

Disasm:


mvirkkunen:
Thanks!

That extra ZLP was definitely an error.

Weird that libusb on Windows doesn't like reading into an empty buffer. Feels like this should be reported on their end too because this is a bit unexpected! Luckily this is just a test program and the fix was simple.

PR #41, opened 2020-09-08
Fixing processing order of EP0 (control) transactions

ryan-summers:
This PR fixes an issue with the processing order of multiple EP0 events.

Specifically, if both EP0-IN and EP0-OUT events have occurred, we **have** to process EP0-IN first so that we update the `ControlState` properly. Otherwise, we inadvertently process the EP0-OUT status phase as a spurious OUT transfer (instead of the STATUS phase as expected).

This only occurs when both EP0-OUT and EP0-IN transactions occur at the same time, which occurs when polling is completed at a lower-than-continuous rate.

After this fix, this crate can be used in `debug` configuration reliably as well.

### DIFF
--- a/src/control_pipe.rs
+++ b/src/control_pipe.rs
@@ -84,6 +84,10 @@ impl<B: UsbBus> ControlPipe<'_, B> {
             },
         };
 
+        // Now that we have properly parsed the setup packet, ensure the end-point is no longer in
+        // a stalled state.
+        self.ep_out.unstall();
+
         /*sprintln!("SETUP {:?} {:?} {:?} req:{} val:{} idx:{} len:{} {:?}",
             req.direction, req.request_type, req.recipient,
             req.request, req.value, req.index, req.length,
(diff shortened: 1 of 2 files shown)


mvirkkunen:
Hmm, yes, I believe processing them in this order will be more reliable because the SETUP is handled second, and that always resets everything back to working order. I wonder if we should also add `ep_out.unstall()` into `ControlPipe` when SETUP is received, just in case the IN handling code still ends up stalling it for some reason?

ryan-summers:
> Hmm, yes, I believe processing them in this order will be more reliable because the SETUP is handled second, and that always resets everything back to working order.

> I wonder if we should also add `ep_out.unstall()` into `ControlPipe` when SETUP is received, just in case the IN handling code still ends up stalling it for some reason?

That doesn't sound like a bad approach to me - I assume we would only `unstall()` when we correctly parse a setup packet.

I think it's also safe to remove the compiler warning when compiling for `debug`.

mvirkkunen:
Thanks! My tests pass for this so I'll release a new version with these changes. This should make things more reliable in the future.

PR #42, opened 2020-09-28
Add basic IAD support

atykhyy:
This PR adds minimal support for composite devices with IADs (interface association descriptors). The motivation is to enable out-of-the-box support of composite devices containing CDC-ACM serial ports and other USB functions (such as HID devices) on Windows.

Some references:
https://www.usb.org/sites/default/files/iadclasscode_r10.pdf
https://docs.microsoft.com/en-us/windows-hardware/drivers/usbcon/usb-interface-association-descriptor
https://www.microchip.com/forums/m189607.aspx
https://community.arm.com/developer/tools-software/tools/f/keil-forum/29075/win7-cdc-hid-only-com-port-works

### DIFF
--- a/src/descriptor.rs
+++ b/src/descriptor.rs
@@ -11,6 +11,7 @@ pub mod descriptor_type {
     pub const STRING: u8 = 3;
     pub const INTERFACE: u8 = 4;
     pub const ENDPOINT: u8 = 5;
+    pub const IAD: u8 = 11;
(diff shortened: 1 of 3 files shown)


mvirkkunen:
Thank you very much! I'm releasing this immediately so people can start using it. The next version of usb-device will have something similar, but with autodetection of composite devices so people don't even have to call a method.

PR #43, opened 2020-10-03
Fix test_class for the High Speed case

Disasm:


Disasm:
There is also a problem with the `bulk_loopback` test: it doesn't take into account the maximum endpoint length in each case, using 64 bytes instead, so for high speed device it fails.

src/test_class.rs
```diff
@@ -83,11 +83,13 @@ impl<B: UsbBus> TestClass<'_, B> {
...
     pub fn make_device<'a, 'b>(&'a self, usb_bus: &'b UsbBusAllocator<B>) -> UsbDevice<'b, B> {
-        UsbDeviceBuilder::new(&usb_bus, UsbVidPid(VID, PID))
+        let builder = UsbDeviceBuilder::new(&usb_bus, UsbVidPid(VID, PID))
             .manufacturer(MANUFACTURER)
             .product(PRODUCT)
-            .serial_number(SERIAL_NUMBER)
-            .build()
+            .serial_number(SERIAL_NUMBER);
+        #[cfg(feature = "test-class-high-speed")]
+        let builder = builder.max_packet_size_0(64);
```
mvirkkunen, line 91: I would put this in `mod sizes` for consistency with the other bits.

Disasm: Than I'm not sure we need it as a constant, unless we want to preserve the default behavior with 8-byte EP0. We can use 64 for both HS and FS.

mvirkkunen: If it's 64 bytes, then the tests in test_class_host have to be changed as well, because they exercise the packet splitting code at different lengths that are under 64 bytes. If the packet size is 64 bytes, multi-packet control transfers wouldn't be tested at the moment. Plus 8 is the default, so I think it'd be a good idea to test with that.

Disasm: Good point, I'll update then!

PR #44, opened 2020-10-04
Add DescriptorWriter::interface_with_string()

Disasm:


Disasm:
I'm not sure about naming and types, maybe we can make it a breaking change and pass `Option<StringIndex>` as `interface()` argument.

mvirkkunen:
Hi, thanks. #55 implements this along with simple alternate setting support, so I'm going to merge that one instead. Sorry for taking so long!

Issue #45, opened 2020-10-17
Difficulty writing descriptors of unknown size


electroCutie:
I'm writing a MIDI descriptor builder for use in another project of mine and realize that it was fairly hard to write the descriptors because they need to know the length of all the MIDI specific headers to put into a Class Specific descriptors

The issue is that the sizes aren't fixed and depend on things such as the number of MIDI entities, the number of virtual "pins" in the topology, and the number of connections in the topology. Counting the number of bytes needed is error prone and making one's own buffer to build on before submitting it to this package is wasteful since there is already a backing buffer in place.

I propose that we make use of the is buffer and expose a lazy form of the writer that will allow descriptors to be written and, at the end, have lengths or other data filled in.

I also have a pull request that implements this idea as a proof of concept that I added to develop my MIDI USB Descriptor builder, so at the very least I can say the idea is sound and I hope it can serve as the basis for an implementation

electroCutie:
Here is an example of how I use the proposed interface in #46 

```
writer.writer(|w| {
    let start_b6 = w.position();
    w.delayed_write(
        1, // outer writes the header size
        |w| {
            w.arr(&[
                constants::DescriptorType::ClassSpecificInterface as u8,
                constants::EndpointSubtype::Header as u8,
                // Revision, 1.0 BCD in Little-endian
                0x00,
                0x01,
            ]);
            //header is the length of that plus the two bytes that describe the total length
            let header_length_b6 = (w.position() - start_b6) + 2;

            // writes the rest and then the length of that into the header
            w.delayed_write(
                2,
                |w| {
                    for e in self.entities {
                        e.write_descriptor(w);
                    }

                    // this length includes the header, we reuse the start from above
                    (w.position() - start_b6) as u16
                },
                |slice, size| slice.copy_from_slice(&size.to_le_bytes()),
            );

            //the outer wrapper writes the header length
            header_length_b6
        },
        |slice, size| slice[0] = size as u8,
    );
})?;
```

electroCutie:
Figured I'd ping this

Any thoughts or feelings on it or the PR?

agausmann:
`usb-device` already does something like this internally to write the number of endpoints in an interface. Because it doesn't know how many endpoints there are when the interface descriptor is written, It writes out a placeholder zero and remembers the position, and then every time `DescriptorWriter.endpoint` is called, it increments the byte at that  position.

[...]

PR #46, opened 2020-10-17
Add a tool for writing descriptors with lengths not known at compile time

electroCutie:
This is meant to be a starting point for #45

### DIFF
--- a/src/descriptor.rs
+++ b/src/descriptor.rs
@@ -43,6 +43,104 @@ pub struct DescriptorWriter<'a> {
     write_iads: bool,
 }
 
+/// Write arbitrary sequences of bytes to descriptors with tools for delayed length writing
+/// If there is an error it is delayed until the end of writing when finish is called
+pub struct ByteWriter<'a, 'b> {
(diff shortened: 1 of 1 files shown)


mvirkkunen:
Hi, thanks for the PR and sorry for the late reply!

I wonder if this couldn't be done a bit more simply with just a few extra methods in `DescriptorWriter`, something along the lines of:

```
    /// Returns a mark for the current byte position
    fn pos(&self) -> Mark { /* ... */ }

    /// Write raw bytes to the descriptor
    fn write_raw(&self, data: &[u8]) -> Result {
        self.write_raw_at(data, self.pos())
    }

    /// Write raw bytes to the descriptor at a previously marked location
    fn write_raw_at(&self, data: &[u8], pos: &Mark) -> Result { /* ... */ }
```

`Mark` would be a simple wrapper type for a `usize` that also supports subtraction for finding out the length of something. This would allow for any weird descriptors that may not encode length as a single byte, or perhaps want to write `length - 1` or something like that.

PR #51, opened 2021-01-23
Support devices with SetAddress handled in hardware (Nordic nRF52) 

unrelentingtech:
Nordic chips handle address setting entirely in hardware so this has to be skipped.
See https://github.com/nrf-rs/nrf-hal/pull/144 for context.


wez:
See also: https://github.com/mvirkkunen/usb-device/issues/4

Disasm:
I implemented this function in the driver itself: https://github.com/nrf-rs/nrf-hal/pull/295/commits/1b3e38b19e035eedd05ed668ad9cb7b2833656c8.
Also I figured out that without this function USB driver works without any problems!

PR #53, opened 2021-03-06
Add imxrt-usbd to the hardware driver crates list

mciantyre:
We're working on a usb-device implementation for NXP's i.MX RT
microcontrollers. As of this writing, the driver passes the usb-device
test suite, and it plays well with usbd-serial. Though the peripheral
supports high speed, we're limiting the driver to full speed, since
the usb-device ecosystem doesn't seem high-speed capable.

Not sure if you're still advertising hardware driver crates. But I
figured I'd add a new item, and take the chance to say thanks for
putting this usb-device ecosystem together.


Disasm:
> Though the peripheral supports high speed, we're limiting the driver to full speed, since the usb-device ecosystem doesn't seem high-speed capable.

We have a proof-of-concept driver implementation for STM32F723 that uses an internal high-speed PHY. No changes to `usb-device` are required to make it work in high-speed-only mode.

PR #54, opened 2021-03-06
Add make_device_builder to test class

mciantyre:
The imxrt-usbd driver requires that the user sets the EP0 max packet
size to 64 bytes. To meet the requirement in our test_class example,
we need access to the UsbDeviceBuilder. Ideally, we could access the
builder without duplicating the make_device method.

This commit adds a new method, make_device_builder, for the test
class. It sets the required device properties, then returns the
builder to the caller for any additional configurations.

### DIFF
--- a/src/test_class.rs
+++ b/src/test_class.rs
@@ -83,11 +83,23 @@ impl<B: UsbBus> TestClass<'_, B> {
 
     /// Convenience method to create a UsbDevice that is configured correctly for TestClass.
     pub fn make_device<'a, 'b>(&'a self, usb_bus: &'b UsbBusAllocator<B>) -> UsbDevice<'b, B> {
+        self.make_device_builder(usb_bus).build()
+    }
+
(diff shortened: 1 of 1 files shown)


Disasm:
> The imxrt-usbd driver requires that the user sets the EP0 max packet size to 64 bytes.

This sounds like a bug. IMXRT1060RM also contains a table (42-42 "Typical Low-/Full-speed Transaction Times") that mentions 8-byte control transactions.

mciantyre:
Thanks for this comment. I was convinced I needed the larger packet size given some earlier experiments. But I tried again, and you're right that it works. I've removed the requirement from `imxrt-usbd`.

This PR may still be useful for anyone who's interested in testing across the allowable EP0 packet sizes. I'll keep this open, but feel free to close if we feel otherwise.

mvirkkunen:
Thanks! Letting people configure the TestClass sure doesn't hurt.

Issue #49, opened 2020-12-19
Alternate setting support


kiffie:
Hi all,

I did not find support for Alternate Settings for Interfaces. Any progress here? As I needed this for one of my projects (USB audio), I use a quite pragmatic solution. Maybe, you would like to integrate it or something similar.

A function like `interface_with_alternate_setting` could be added to `descriptor.rs`, which must be called multiple times to create multiple interface descriptors for the respective alternate settings. The SET_INTERFACE request can then be handled in an ` impl<B: UsbBus> UsbClass<B>`. Later, the request handler in `device.rs` could be extended to avoid boilerplate code in the UsbClass impl.

In case you are interested in this approach of adding alternate settings, I could create a PR.

Here is the code of the the function I added to my own copy of your code during my experiments:

```rust
    /// Writes a interface descriptor with a specific alternate setting.
    ///
    /// # Arguments
    ///
    /// * `number` - Interface number previously allocated with
    ///   [`UsbBusAllocator::interface`](crate::bus::UsbBusAllocator::interface).
    /// * `alternate_setting` - number of the alternate setting
    /// * `interface_class` - Class code assigned by USB.org. Use `0xff` for vendor-specific devices
    ///   that do not conform to any class.
    /// * `interface_sub_class` - Sub-class code. Depends on class.
    /// * `interface_protocol` - Protocol code. Depends on class and sub-class.
    pub fn interface_with_alternate_setting(&mut self, number: InterfaceNumber,
        alternate_setting: u8,
        interface_class: u8, interface_sub_class: u8, interface_protocol: u8) -> Result<()>
    {
        if alternate_setting == device::DEFAULT_ALTERNATE_SETTING {
            match self.num_interfaces_mark {
                Some(mark) => self.buf[mark] += 1,
                None => return Err(UsbError::InvalidState),
            };
        }

        self.num_endpoints_mark = Some(self.position + 4);

        self.write(
            descriptor_type::INTERFACE,
            &[
                number.into(), // bInterfaceNumber
                alternate_setting, // bAlternateSetting (how to even handle these...)
                0, // bNumEndpoints
                interface_class, // bInterfaceClass
                interface_sub_class, // bInterfaceSubClass
                interface_protocol, // bInterfaceProtocol
                0, // iInterface
            ])?;

        Ok(())
    }
```

Best regards,

Stephan

redpfire:
This looks very neat. In my dfu-bootloader project i did something similar:
```rust
[...]

PR #55, opened 2021-03-10
add support for alternate settings

kiffie:
Here is a PR based on the two code snippets from issue "Alternate setting support #49" from @redpfire and myself. @ianrrees, you might want to have a look at it.

 @mvirkkunen, what do you think about it? Your comments are appreciated. If you could merge it, we could get rid of all the personal forks.

### DIFF
--- a/src/descriptor.rs
+++ b/src/descriptor.rs
@@ -1,5 +1,5 @@
 use crate::{Result, UsbError};
-use crate::bus::{UsbBus, InterfaceNumber};
+use crate::bus::{UsbBus, InterfaceNumber, StringIndex};
 use crate::device;
 use crate::endpoint::{Endpoint, EndpointDirection};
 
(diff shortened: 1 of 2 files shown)


src/descriptor.rs
```diff
@@ -175,24 +175,64 @@ impl DescriptorWriter<'_> {
...
+    /// * `number` - Interface number previously allocated with
+    ///   [`UsbBusAllocator::interface`](crate::bus::UsbBusAllocator::interface).
+    /// * `alternate_setting` - number of the alternate setting
+    /// * `interface_class` - Class code assigned by USB.org. Use `0xff` for vendor-specific devices
+    ///   that do not conform to any class.
+    /// * `interface_sub_class` - Sub-class code. Depends on class.
+    /// * `interface_protocol` - Protocol code. Depends on class and sub-class.
+    /// * `interface_string` - Index of string descriptor describing this interface
+
+    pub fn interface2(
```
ianrrees, line 202: I'd prefer a more descriptive name, maybe `interface_alternate()`?

redpfire: Right, a more descriptive name than `interface2` would be fitting.

kiffie: What about the original one: ```interface_with_alternate_setting```?

redpfire: Sounds good but it may be too long.

redpfire: Sounds good to me. Ian?

kiffie: I change the code. What do you mean with 'lan'?

redpfire: What i meant is if Ian has any suggestions. Anyway the change sounds good to me :)

redpfire:
Could you please squash those two commits into one? It will be cleaner after the merge.

kiffie:
Done.

ianrrees:
We should bump the minor version in Cargo.toml as well, since this adds to the API.

redpfire:
Let's bring @mvirkkunen into the discussion.

src/descriptor.rs
mvirkkunen: `interface_ex`? Too Win32?

mvirkkunen:
I think this is an OK solution with the current crate version. It should be sufficient for the most important usecases (DFU, USB audio).

src/descriptor.rs
mvirkkunen: Perhaps even `interface_alt` :thinking:

PR #56, opened 2021-03-14
Add an invite in the README

stappersg:
Removed reservations.
The idea is transmitting:  You  are invited to come with merge request
(The idea is stopping to transmit: I do it all by myself and these are my personal priorities)

Note: That only text is removed, if there was an order in priority, then it is still there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,12 +38,9 @@ Class crates
 TODO
 ----
 
-Features planned but not implemented yet:
+Features
 
 - Interface alternate settings
 - Multilingual string descriptors
 - Isochronous endpoints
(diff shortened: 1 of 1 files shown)


mvirkkunen:
Thanks for the PR. I think it would be best to just remove that section. It's outdated anyways and we have Github issues for things people want/need. Do you want to update the PR or should I close and do it myself?

(From the title I thought you wanted to add a Matrix chat invite, heh)

stappersg:
> Do you want to update the PR or should I close and do it myself?

New PR in preparation.

PR #57, opened 2021-03-14
Removed TODO from README

stappersg:
I have been told it is outdated.


mvirkkunen:
Thanks!

stappersg:
OK

( Today I Learned:    Use also git branches for the simplest changes. It will prevent commit message noise, new branches allow to provide clean merge requests)

mvirkkunen:
Yep, it's best to create a new branch for every pull request. Otherwise if you create two at the same time you'll also easily end up updating both accidentally. I squashed the commits for you so no worries!

PR #60, opened 2021-03-24
Add enumerations and allocator for isochronous endpoints

ianrrees:
Aim is to resolve https://github.com/mvirkkunen/usb-device/issues/33, however as @lkolbly points out it doesn't add testing support.  @mvirkkunen is that OK with you, or should I look in to adding test support?  I expect any practical use of isochronous endpoints to not use the current endpoint buffering arrangement (the `copy_to_nonoverlapping()` takes a fair bit of processor time), so a test that transfers data in/out of an isochronous endpoint would currently be a bit artificial.

This will also need a minor version bump, assuming we're following semver.

@sibiryakov - any comments?

### DIFF
--- a/src/bus.rs
+++ b/src/bus.rs
@@ -1,4 +1,7 @@
-use crate::endpoint::{Endpoint, EndpointAddress, EndpointDirection, EndpointType};
+use crate::endpoint::{
+    Endpoint, EndpointAddress, EndpointDirection, EndpointType, IsochronousSynchronizationType,
+    IsochronousUsageType,
+};
 use crate::{Result, UsbDirection, UsbError};
 use core::cell::RefCell;
 use core::mem;
@@ -244,6 +247,38 @@ impl<B: UsbBus> UsbBusAllocator<B> {
             .expect("alloc_ep failed")
     }
 
+    /// Allocates an isochronous endpoint.
+    ///
+    /// # Arguments
+    ///
+    /// * `synchronization` - Type of synchronization used by the endpoint
+    /// * `usage` - Whether the endpoint is data, explicit feedback, or data+implicit feedback
+    /// * `payload_size` - Payload size in bytes.
+    /// * `interval` - Interval for polling, expressed in frames/microframes.
+    ///
+    /// See USB 2.0 section 9.6.6.
+    ///
+    /// # Panics
+    ///
+    /// Panics if endpoint allocation fails, because running out of endpoints or memory is not
+    /// feasibly recoverable.
+    #[inline]
+    pub fn isochronous<D: EndpointDirection>(
+        &self,
+        synchronization: IsochronousSynchronizationType,
+        usage: IsochronousUsageType,
+        payload_size: u16,
+        interval: u8,
+    ) -> Endpoint<'_, B, D> {
+        self.alloc(
+            None,
+            EndpointType::Isochronous((synchronization, usage)),
+            payload_size,
+            interval,
+        )
+        .expect("alloc_ep failed")
+    }
+
     /// Allocates a bulk endpoint.
     ///
     /// # Arguments
@@ -263,6 +298,7 @@ impl<B: UsbBus> UsbBusAllocator<B> {
     /// Allocates an interrupt endpoint.
     ///
     /// * `max_packet_size` - Maximum packet size in bytes. Cannot exceed 64 bytes.
+    /// * `interval` - Polling interval.
     ///
     /// # Panics
     ///
--- a/src/bus.rs
+++ b/src/bus.rs
@@ -272,7 +272,10 @@ impl<B: UsbBus> UsbBusAllocator<B> {
     ) -> Endpoint<'_, B, D> {
         self.alloc(
             None,
-            EndpointType::Isochronous((synchronization, usage)),
+            EndpointType::Isochronous {
+                synchronization,
+                usage,
+            },
             payload_size,
             interval,
         )
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+* New enums and allocators for Isochronous endpoints
 
-...
+### Changed
+* `EndpointType` enum now has fields for isochronous synchronization and usage.
 
 ## [0.2.9] - 2022-08-02
 
--- a/src/descriptor.rs
+++ b/src/descriptor.rs
@@ -324,7 +324,7 @@ impl DescriptorWriter<'_> {
             let mps = endpoint.max_packet_size();
 
             buf[0] = endpoint.address().into();
-            buf[1] = endpoint.ep_type() as u8;
+            buf[1] = endpoint.ep_type().to_bm_attributes();
             buf[2] = mps as u8;
             buf[3] = (mps >> 8) as u8;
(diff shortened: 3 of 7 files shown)


dgoodlad:
@ianrrees I've been working on support on STM32 chips over in stm32-rs/synopsys-usb-otg#29. Did you ever get feedback on this? Is the only real blocker testing support?

ianrrees:
Hi @dgoodlad, from across the Tasman!  I've not had more feedback about this, but it's looking like I will be interested in it for some USB audio work sometime in the next several months...

Haven't had a chance to look in to testing support; it seems like something that would be helpful but perhaps a simplified "unit test" approach would be easy without needing to implement isochronous transfers in the test harness.

I'll aim to update the PR today anyway.

dgoodlad:
That'd be amazing! I'm unfamiliar with how the test harness etc works here, so got a bit stuck as to how to approach it myself.

I've been hacking on some USB audio stuff over [here](https://github.com/dgoodlad/crabdac), including the basics of USB Audio Class 2. That's what I've been using to drive out support in the other linked PR, so I know that it at least works for isochronous IN endpoints.

In that work, it did become clear that it would be good to have some extra events on the usb bus & device to allow implementers to handle incomplete transfers properly. If this one gets merged I'd be happy to follow up with a PR suggesting changes that felt somewhat ergonomic when I was hacking away :)

ianrrees:
I've rebased on to current master, though didn't test it against my project (would require a fair bit of work).  That said, it's not a very big change and I can't see any reason that the rebase would've introduced new problems.

Also looked a little at testing, but I must admit my memory is a bit fuzzy about the details.  Since this PR is really just about setting the right attributes bits in endpoint descriptors, test coverage for this PR doesn't actually require doing a /transfer/.  Isochronous transfers were the main sticking point in implementing tests, from memory.  At least at the time, `rusb` didn't support `libusb`'s asynchronous interface, which is needed to do isochronous transfers.  So, it might not be that hard to add tests to this PR, if we're OK with not actually testing isochronous transfers?

How does the testing work at an organisational level?  It looks like actually running the tests requires someone to run them on real hardware.

eldruin:
Maybe @mvirkkunen can clarify here how to run these tests.

Disasm:
Currently tests can be executed manually by flashing firmware with `test_class` from this repo to a device and running `cargo test` with the device connected to USB.
Test class code for STM32F446: https://github.com/Disasm/usb-otg-workspace/blob/master/example-f446ze-board/examples/test_class.rs

ianrrees:
Looking at this again with a fresh set of eyes (and more Rust experience), I'm not sure that the new types are actually necessary and I don't like the tuple-style enum...  Please hold.

dgoodlad:
> Looking at this again with a fresh set of eyes (and more Rust experience), I'm not sure that the new types are actually necessary and I don't like the tuple-style enum... Please hold.

I like this a lot better 👍

ianrrees:
Sweet, I'll try to bang together a test setup this weekend (it'll be SAMD-based) and add test coverage for this.

Testing isochronous transfers thoroughly might actually be a hard problem, since they're not supposed to be reliable, and TBH I think it's a bit outside the scope of this crate anyway.

ianrrees:
I've made a start at testing, but think there might be something wrong with either the ATSAMD HAL or my hardware - haven't managed to get this board to enumerate...  Might try later today with a board that uses a different chip that I know the HAL USB stuff works with, but have some IRL things to attend to.

Work-in-progress is here:

https://github.com/ianrrees/test-usb-device

ianrrees:
Ok, new tests are implemented and confirmed to pass on ATSAMD21 against [this branch](https://github.com/ianrrees/atsamd/tree/isochronous-usb) of the atsamd HAL.

I'll open up an issue to discuss that test-usb-device repo - maybe it's something that we should have in this project or an adjacent one.

Cryowatt:
Is there something else that needs to be done here to get this integrated? Even if testing is a blocker, I'd like to see this shipped as an unstable feature at the very least.

ianrrees:
This PR adds test coverage that I'd say is on par with the rest of usb-device.  My vague memory is that in August we wound up talking about a general level-up in USB testing, but that's a bit beyond the scope of this PR.  LMK if there's anything I can do to help this along, but it'll be a while before I'll have a significant amount of time for work on Rust USB stuff.

ryan-summers:
@ianrrees If you have this tested and somewhat operational, I'm fine merging this as-is without the additional testing. Just let me know if it's in a usable state

src/bus.rs
```diff
@@ -244,6 +247,40 @@ impl<B: UsbBus> UsbBusAllocator<B> {
...
+    ///
+    /// Panics if endpoint allocation fails, because running out of endpoints or memory is not
+    /// feasibly recoverable.
+    #[inline]
+    pub fn isochronous<D: EndpointDirection>(
+        &self,
+        synchronization: IsochronousSynchronizationType,
+        usage: IsochronousUsageType,
+        payload_size: u16,
+        interval: u8,
```
ryan-summers, line 270: Can we do input validation on the `interval` and `payload_size` to verify they conform with USB requirements?

src/bus.rs
```diff
@@ -244,6 +247,40 @@ impl<B: UsbBus> UsbBusAllocator<B> {
...
+        self.alloc(
+            None,
+            EndpointType::Isochronous {
+                synchronization,
+                usage,
+            },
+            payload_size,
+            interval,
+        )
+        .expect("alloc_ep failed")
```
ryan-summers, line 281: I would just pop the error out as a `Result<>` instead of unwrapping with the `expect()` call

ryan-summers (changes requested):
Some very minor nits around unwrap/expect and input validation, but everything looks fine otherwise

ianrrees:
> Some very minor nits around unwrap/expect and input validation, but everything looks fine otherwise

Thanks!  I'll try to make time for this in the next few days.  It's been a while since I've had anything to do with this stuff so just want to get re-oriented before responding.

src/bus.rs
ianrrees: I'm in favour of this in principle, however the other allocator methods use `.expect()`  and I think it's probably better to either stick with that convention or change all of the allocators to return `Result<>`.  The second would be a breaking change, so might be better combined with other big-ish changes?

src/bus.rs
ianrrees: Ah, duh, I was thinking about https://github.com/rust-embedded-community/usb-device/issues/71

src/bus.rs
ryan-summers: I personally dislike deferring to current design, but I think you're right in this case that we don't gain anything via changing this expect (i.e. it does not remove format bloat).

We should spawn an issue to handle converting to `Result<>` return types to avoid format bloat

ianrrees: I like the idea of splitting that issue out.  It might be a good improvement to combine with some more flexible memory allocation scheme; IIRC the current one (or maybe just be the ATSAMD implementation) just uses a fixed-size buffer, which as the comment says is a bit limiting.

src/bus.rs
ianrrees: Had a closer look at this tonight, I don't think we have quite enough information to validate at this level.  Might be able to do some of the fields, but others need to know about the USB version and speed of the device.  As in #71 I do think it would be appropriate to do in this crate, just a bit outside the scope of this PR.

ianrrees:
Have rebased on to current master and touched up a few formatting/comment issue.  I'm in favour of merging this as-is, and capturing those other improvements in the bug tracker as I do think they'd be nice to have eventually.

ryan-summers (approved):

PR #63, opened 2021-04-12
Add write_iter and endpoint_with_additional_data methods

Windfisch:
Hi,

USB Audio Class devices (including MIDI devices) extend the endpoint descriptor by two additional bytes, *bRefresh* and *bSynchAddress* (c.f. [the USB MIDI 1.0 spec](https://usb.org/sites/default/files/midi10.pdf) on page 25/26).

Currently, usb-device offers no way to create such an endpoint. I've added such a function and also a more generic way to `write()` into the descriptor: `write_iter` accepts an iterator instead of a slice, allowing for more flexibility using e.g. chained iterators.

I would be happy if this could be merged as it should not cause any incompatibilities.

It would be possible to replace the `endpoint()` method's implementation body by just a call to `endpoint_with_additional_data()` with an empty iterator at a tiny runtime cost.

If you prefer to have this change as well or have anything else to discuss, please let me know!

### DIFF
--- a/src/descriptor.rs
+++ b/src/descriptor.rs
@@ -79,6 +79,28 @@ impl DescriptorWriter<'_> {
         Ok(())
     }
 
+    /// Writes an arbitrary (usually class-specific) descriptor.
+    pub fn write_iter(&mut self, descriptor_type: u8, descriptor: impl IntoIterator<Item=u8>) -> Result<()> {
+        let start = self.position + 2;
+
+        let mut length = 0;
+        for byte in descriptor {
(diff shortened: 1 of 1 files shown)


JosefUtbult:
@Windfisch Hi there. I've been working on trying to get an asynchronous endpoint to work for my soundcard, but I ran into the same problem as you with the lack of fields in the endpoint descriptor. I've been looking through your pull request, but I have a hard time understanding how it should be applied when creating an endpoint. Would you mind helping me? Thanks in advance

PR #64, opened 2021-04-27
Add methods to make descriptor writing more flexible

mvirkkunen:
Adds a method for writing descriptors with an unknown length in a way that's similar to how accepting control transfers works, and a method for writing endpoint descriptors with extra data in the same manner.

Unfortunately the naming is the reverse of control transfers (`write_with` takes a callback while `accept_with` takes a buffer), but this way it's compatible.

This would implement both #63 and #46. @Windfisch, if you have an audio driver going, could you test this to see if it's good for your purposes?

### DIFF
--- a/src/descriptor.rs
+++ b/src/descriptor.rs
@@ -1,3 +1,5 @@
+use core::cmp::min;
+
 use crate::bus::{InterfaceNumber, StringIndex, UsbBus};
 use crate::device;
 use crate::endpoint::{Endpoint, EndpointDirection};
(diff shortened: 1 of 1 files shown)


Windfisch:
Hey @mvirkkunen, thank you for your effort!
This PR works fine for me (though I still need to manually merge in #62 to support larger descriptors ;)).
I also think that the interface is nice to use, so I'm fine :).

Windfisch:
@mvirkkunen care to click "merge" on this?

PR #67, opened 2021-05-09
Adding basic remote wakeup hooks

### LINKED ISSUE #65: Resume (remote wakeup) hooks
I see a bunch of places in the code to enable "remote_wakeup". However, I can seem to find where to trigger the remote wakeup sequence from the perspective of the UsbBus trait.

I'm currently implementing this hal for atsam4 (which is very different from atsamd) and should hopefully get some code pushed in the next week or so (I have ep0 up and running now!).

In the context of a keyboard, when a key is pressed and the computer is asleep you need to check if the USB module is in a suspend state. 

> In Suspend state it is possible to wake up the host sending an external resume. 
> - The device must wait at least 5 ms after being entered in suspend before sending an external resume.
> - The device has 10 ms from the moment it starts to drain current and it forces a K state to resume the host.
> - The device must force a K state from 1 to 15 ms to resume the host

In my case for atsam4s I need to set a bit in the USB (UDP) register block, then clear it immediately after.
I know from other MCUs (mk20dx256vlh7 iirc), you actually need to wait for an interrupt on when to clear this bit in order to handle the 1 to 15 ms timing. Other MCUs you have to keep track of this yourself.

Have these cases been considered in this crate yet?

haata:
- Fixes #65
- Adds a new error type to indicate that the bus is not suspended
  (Bus must be suspended to initiate Remote Wakeup)
- The poll functions can be used for implementations that require
  interrupt handling after initiating the remote wakeup

### DIFF
--- a/src/bus.rs
+++ b/src/bus.rs
@@ -120,6 +120,21 @@ pub trait UsbBus: Sync + Sized {
     /// interrupt handler. See the [`PollResult`] struct for more information.
     fn poll(&self) -> PollResult;
 
+    /// Initiates the remote wakeup sequence
+    /// Implementation can vary wildly as some hardware handles the timing, others do not.
+    ///
+    /// The default implementation just returns `Unsupported`.
+    ///
+    /// # Errors
+    ///
+    /// * [`Unsupported`](crate::UsbError::Unsupported) - This UsbBus implementation doesn't support
+    ///   remote wakeup.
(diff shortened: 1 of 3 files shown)


haata:
Any thoughts on this PR?

haata:
Sorry for the delay, I went ahead and implemented your suggestion (it's working! and passes the USB compliance tool too!).

The main annoyance (and I'm not really sure how to get around this), is using remote_wakeup_enabled flag inside the USB driver. Just a few more lines of setup code, so not really a big deal.

PR #68, opened 2021-05-09
Fixixing most of the clippy warnings

haata:


haata:
https://github.com/mvirkkunen/usb-device/pull/75 for full set of fixes.

PR #69, opened 2021-05-09
Make USB device config public

nickray:
I was about to start tackling USB/IP when I discovered @Sawchord's https://github.com/Sawchord/usbip-device.

For `usbip list -r localhost` to be somewhat pretty, the USB/IP server should return basic info on the device, which duplicates the contents of the non-public `Config` struct here. (It's not strictly necessary for this info to be correct, AFAIU.)

Looking at UsbDeviceBuilder, in reality it's a ConfigBuilder, which would become quite redundant (candidate for deprecation?) if the config itself were made public with a fluent/builder interface. I'm gambling on increasing merge/publish potential by staying backwards compatible though :)

This would then allow re-use in the USB/IP driver (e.g. https://github.com/nickray/usbip-device/blob/7bab96f64072da82a552c94213b5dc98f2481c99/examples/twitching_mouse.rs#L16-L26)

There's a bit chicken+egg involved regarding DRY and the interface descriptors (which USB/IP also advertises); solving that seems it would require breaking changes which I don't want to propose.

P.S. Are there any/many useful cases for `Config<'a>` with non-`'static` lifetimes?

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,52 +1,72 @@
-use crate::bus::{UsbBus, UsbBusAllocator};
-pub use crate::config::UsbVidPid;
-use crate::{config::Config, device::UsbDevice};
-
-/// Used to build new [`UsbDevice`]s.
-pub struct UsbDeviceBuilder<'a, B: UsbBus> {
-    alloc: &'a UsbBusAllocator<B>,
-    config: Config<'a>,
-}
+/// A USB vendor ID and product ID pair.
+pub struct UsbVidPid(pub u16, pub u16);
 
 macro_rules! builder_fields {
     ( $( $(#[$meta:meta])* $name:ident: $type:ty, )* ) => {
         $(
             $(#[$meta])*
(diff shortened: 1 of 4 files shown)


src/config.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub serial_number: Option<&'a str>,
+    /// Is device self-powered?
+    pub self_powered: bool,
+    /// Does device support remote wakeup?
+    pub supports_remote_wakeup: bool,
+    /// Is device composite with IADs?
+    pub composite_with_iads: bool,
+    /// Maximum power consumption of device
+    pub max_power: u8,
+}
```
ryan-summers, line 47: As much as I like the builder semantics, making struct fields public is problematic . See https://rust-lang.github.io/api-guidelines/future-proofing.html#structs-have-private-fields-c-struct-private for a discussion on why this is problematic for a maintainability perspective.

Other than that, I like the builder config API

PR #70, opened 2021-05-18
Make Send+Sync impls optional for compatibility with AVR

agausmann:
The AVR architecture doesn't have atomic operations, but the chips are generally single-core anyway so thread safety isn't necessary.

This crate can't be used on AVR in official Rust yet, since this crate uses `dyn` / trait objects (rust-lang/rust#79889). But there is a pending fix, and I'm testing it out with [my own patched version of Rust/LLVM](https://github.com/agausmann/rust/tree/avr-patch/).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,6 @@ control-buffer-256 = []
 # speed device, but the descriptors will be invalid.
 test-class-high-speed = []
 
-
 [[test]]
 name = "test_class_host"
 path = "tests/test_class_host/main.rs"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,11 +14,9 @@ rusb = "0.8.0"
 rand = "0.6.1"
 
 [features]
-default = ["sync"]
-
 # Enable support for using this library across multiple threads.
 #
-# Enabled by default, but made optional for compatibility with AVR targets
+# Made optional / opt-in for compatibility with AVR targets
 # which are single-core and do not have atomic operations.
 sync = []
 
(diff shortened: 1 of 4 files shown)


ianrrees:
Hi @agausmann , it's cool to see Rust coming to AVR!  The thread safety actually is still necessary on single-core chips, because interrupts provide additional execution contexts.  The ATSAMD21 for instance also lacks a compare-and-swap (CAS) instruction, the usb-device implementation in the HAL uses an interrupt-free block ([here](https://docs.rs/cortex-m/0.7.0/src/cortex_m/interrupt.rs.html#55-73)) to implement a "mutex".

However, that approach can be problematic, as I ramble on about in [this issue](https://github.com/atsamd-rs/atsamd/issues/397#issuecomment-861098627).

Rahix:
@agausmann, you can make "atomics" work on AVR by setting

```json
	"max-atomic-width": 16,
```

in your rustc-target JSON.  This works because there are intrinsics that implement the atomic operations by disabling interrupts for their duration.  However, I am not certain that they are always compiled correctly at this time.

agausmann:
Sorry, I meant to respond to this a long time ago. Just want to clear some things up:

>  The thread safety actually is still necessary on single-core chips, because interrupts provide additional execution contexts.

Okay, so my comment about "chips are single-core anyway" wasn't accurate in hindsight. Don't even know why I wrote that.

What I should have said: The motivation for this is to use (or at least try to use) usb-device only on the main execution thread without accessing it in ISRs; I've had success with that approach on other MCUs, and I wanted to try that on AVR. This PR was a potential solution for that, I wanted to opt-in to !Send + !Sync to work around the lack of atomics in AVR. (also thanks @Rahix for helping with that)

agausmann:
No longer interested in this

PR #73, opened 2021-08-06
USB 2.0 HS

mriise:
This is the beginnings of HS the protocol. 

currently only makes interval generic to allow for Microseconds when the time comes.


mriise:
For speed configuration I am currently thinking something along the lines of
```rust
pub enum UsbSpeed {
    Usb,
    Usb2(Usb2Speed),
}

pub enum Usb2Speed {
    FullSpeed,
    HighSpeed,
}
```

Usb 3.0 could be possible, but I am focusing on just getting high speed for this PR.

PR #76, opened 2021-08-23
Defmt support for enums and structs

haata:
- Hidden behind defmt-impl feature
  (similar to how heapless handles defmt)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,9 @@ license = "MIT"
 authors = ["Matti Virkkunen <mvirkkunen@gmail.com>"]
 repository = "https://github.com/mvirkkunen/usb-device"
 
+[dependencies]
+defmt = { version = "0.3", optional = true }
+
 [dev-dependencies]
 rusb = "0.8.0"
 rand = "0.6.1"
--- a/src/bus.rs
+++ b/src/bus.rs
@@ -281,6 +281,7 @@ impl<B: UsbBus> UsbBusAllocator<B> {
 
 /// A handle for a USB interface that contains its number.
 #[derive(Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct InterfaceNumber(u8);
 
 impl From<InterfaceNumber> for u8 {
@@ -291,6 +292,7 @@ impl From<InterfaceNumber> for u8 {
 
 /// A handle for a USB string descriptor that contains its index.
 #[derive(Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "defmt", derive(defmt::Format))]
 pub struct StringIndex(u8);
 
 impl StringIndex {
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -42,3 +42,8 @@ jobs:
         with:
           command: check
           args: --features control-buffer-256
+
+      - uses: actions-rs/cargo@v1
+        with:
+          command: check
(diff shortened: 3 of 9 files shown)


stappersg:
But why?


Longer version:
The pull request has been seen, not yet seen by me is what it improves. Consider that it is me who isn't seen the larger picture. Please elaborate _defmt_ and tell what it makes better.

haata:
USB tends to be timing sensitive, so in order to add logging inside a USB hal crate (what I'm currently working on https://github.com/atsam-rs/atsam4-hal/pull/49/files) defmt is necessary.

My current development flow is to use probe-run + defmt for logging and bracktrace support.
defmt (https://ferrous-systems.com/blog/defmt/ is a good read, but I'll give the tl;dr here) strips out debug messages from the firmware, places it on the host side, then uses a protobuf-like mechanism to pass any necessary arguments back to the host to reassemble the print message. The benefit, significantly smaller firmware images (if you have lots of messages) as well as faster logging as fewer bytes need to be sent over the debug link (e.g. RTT).

If these derive macros aren't added, I would need to maintain either a fork or wrap each enum and struct with another struct in order to handle log message formatting.

Does that make sense?

stappersg:
On Tue, Aug 24, 2021 at 12:45:19AM -0700, Jacob Alexander wrote:
> USB tends to be timing sensitive, so in order to add logging inside a USB hal crate  defmt is necessary.
> 
>    ... further good text ... 
>  ...(https://ferrous-systems.com/blog/defmt/ is a good read ...
> 
> Does that make sense?


Yes, it does.


It's worth to add to the commit message.

haata:
Sorry for the delay, I've been under the weather.

Cargo.toml
```diff
@@ -24,6 +27,9 @@ control-buffer-256 = []
 # speed device, but the descriptors will be invalid.
 test-class-high-speed = []
 
+# Add defmt Format support for enums and structs
+defmt-impl = ["defmt"]
```
ryan-summers, line 31: Let's copy how `smoltcp` does this. I don't think we need to add an explicit feature flag. The user can just do `features=["defmt"]` and that should be sufficient.
https://github.com/smoltcp-rs/smoltcp/blob/cae682e0885502cc3fa4e122d44f67167addb095/Cargo.toml#L24

https://github.com/smoltcp-rs/smoltcp/blob/cae682e0885502cc3fa4e122d44f67167addb095/src/socket/icmp.rs#L25

ryan-summers (changes requested):
Can you please also add a CHANGELOG entry for adding defmt implementations for various structures?

haata:
Done

src/bus.rs
```diff
@@ -281,6 +281,7 @@ impl<B: UsbBus> UsbBusAllocator<B> {
 
 /// A handle for a USB interface that contains its number.
 #[derive(Copy, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "defmt-impl", derive(defmt::Format))]
```
ryan-summers, line 284: All of the code blocks use feature `defmt-impl` instead of just `defmt`

ryan-summers (changes requested):

src/bus.rs
haata: Added another cargo check with the defmt feature.

haata: Good catch! It's been a while since I wrote the PR.

haata:
Thanks!

ryan-summers:
No problem - thanks for the help! :)

PR #78, opened 2021-12-09
add maybe_write() for lazy data closures on Endpoint and UsbBus

mutantbob:
The `write()` API for `Endpoint` is a little clumsy because it returns an error if the write would block, forcing the code that invokes it to hang on to the buffer and retry later.

This is *a* solution, but it adds method to the `UsbBus` trait, forcing downstream implementers to update.

I am mostly trying to start a conversation about this so options (such as the ability to check `WouldBlock` before calling `write`)  can be discussed.

### DIFF
--- a/src/bus.rs
+++ b/src/bus.rs
@@ -78,6 +78,29 @@ pub trait UsbBus: Sync + Sized {
     /// Implementations may also return other errors if applicable.
     fn write(&self, ep_addr: EndpointAddress, buf: &[u8]) -> Result<usize>;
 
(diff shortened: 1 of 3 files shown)


mutantbob:
An example of how it might simplify code that uses USB: https://github.com/mutantbob/imxrt-usbd/commit/05c65845587a61b3bb311eced137a9a0560a6a20

ithinuel:
Wouldn't an `fn is_writable(&self) -> bool` method be simpler?

PR #83, opened 2022-01-18
Adding handling of short DATA stage during SETUP transactions

### LINKED ISSUE #82: USB device may fail initial enumeration on Windows
### Overview
During development, I noted that my USB CDC-ACM virtual serial port was enumerating on Ubuntu without issue, but the enumeration process on Windows was failing with a note "Set Address failed".

This occurred in both debug and release configurations.

### Analysis
To debug, I added RTT logging of EP0 control states, inbound requests, and written data and collected logs from the firmware on both Windows and Ubuntu.

Analysis of the logs indicates that both operating systems conduct what appears to be a partial read of the `DEVICE_DESCRIPTOR` as the very first communication with the device. The descriptor is 18 bytes long, but the OS appears to end the DATA stage of the control transfer early after receiving the first 8 or 16 bytes (presumably because it is only interested in determining the max packet size of EP0).

On Ubuntu, the device logs a RESET after the partial descriptor read, but this reset is not present on Windows.

After the partial descriptor read, EP0 then receives a `SET_ADDRESS` configuration request.

On Windows, this is where the configuration process halted. On Ubuntu, the process continues normally. My current assumption is that the `usb-device` goes into an error state and stalls one of the EP0 IN/OUT channels after the partial descriptor read, which causes the SET_ADDRESS to fail. On Ubuntu, a RESET occurs which removes the stall condition.

ryan-summers:
This PR fixes #82 by allowing a EP0-OUT data packet reception during a DATA-IN transmission to the host. The host may not read the entire descriptor and end the DATA phase early. In these cases, we were stalling EP0, which resulted in the next stage failing.

This PR updates it so that the EP0 control state is simply transitioned to IDLE and the DATA phase is truncated, as the host prematurely ended the transfer.


yorickdewid:
I've tested the PR and it does fix the issue. Windows now enlists the USB device.

ryan-summers:
@eldruin Would you mind taking a look at this PR? I don't want to approve and merge my own changes, but I've been using these changes successfully for quite a while.

eldruin (approved):
Looks good to me. Thank you!

PR #97, opened 2022-06-10
Fix Suspend state transition to go back to the previous state (#96)

### LINKED ISSUE #96: Why does Suspend always transition to Default on Resume?
I noticed an oddity testing remote wakeup where if another device does the wakeup and the device under test resumes usb-device crate will set the state to UsbDeviceState::Default rather than the previous state.

My general understanding of the state transitions follows (https://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-11100-32-bit%20Cortex-M4-Microcontroller-SAM4S_Datasheet.pdf; 40.6.3 page 1042).
screenshot7

This diagram implies that when transitioning back from Suspend, you should go back to the previous state, rather than Default state.

https://github.com/rust-embedded-community/usb-device/blob/413b77cd38d200c3b0e777dd3055cb7ef82c46a0/src/device.rs#L162-L175

Is this intended? Should I make a PR to fix it?

haata:
- Per (https://ww1.microchip.com/downloads/en/DeviceDoc/Atmel-11100-32-bit%20Cortex-M4-Microcontroller-SAM4S_Datasheet.pdf; 40.6.3 page 1042), the state transition from Suspend should go back to the previous state
- Previously Suspend would always transition back to Default state
- Fixes #96

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 * Fixed an issue where USB devices were not enumerating on Windows ([#32](https://github.com/rust-embedded-community/usb-device/issues/82))
 * Add optional support for defmt ([#76](https://github.com/rust-embedded-community/usb-device/pull/76))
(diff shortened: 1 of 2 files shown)


eldruin (approved):
Good catch, thank you!

Issue #100, opened 2022-08-06
How to manage testing usb-device?


ianrrees:
For [this PR](https://github.com/rust-embedded-community/usb-device/pull/60) I've thrown together a [firmware and minimal documentation](https://github.com/ianrrees/test-usb-device) to use this crate's `TestClass`.  @Disasm has an implementation for a different micro [here](https://github.com/Disasm/usb-otg-workspace/blob/master/example-f446ze-board/examples/test_class.rs).

I wonder if we should associate test firmware sources with this project somehow?  It's not clear to me whether that should look more like a link in usb-device documentation, checking test firmware sources in to usb-device, or putting the test firmwares with HALs that implement usb-device support.  Any thoughts?

For me, this was a bit of a mission (partly down to flaky SWD, it seems), and I think it would be good to reduce friction around testing as much as practical.

Disasm:
There is also this repo for a different driver: https://github.com/Disasm/stm32-usbd-tests
I think a collection of links in this repo with some test setup description would be good to have.

ianrrees:
Ah, cool, thanks @Disasm .  I'll open a PR to the main usb-device README.md as a start.  Do you have links to boards that someone interested in testing STM32s might be able to buy?

Does the USB2 test suite pass?  I've not got a login handy but could have a look at that at work sometime.

dgoodlad:
> Do you have links to boards that someone interested in testing STM32s might be able to buy?

The various STM32 "Discovery" boards are high quality and include a debugger compatible with `probe-run`. Other cheaper options are e.g. the cheap [Black Pill](https://github.com/WeActStudio/WeActStudio.MiniSTM32F4x1) boards based on the STM32F4x1 chips.

ianrrees:
Following on from discussion in PRs #101 and #103 - my feeling is that the hardware specific code+documentation should not live in `usb-device`, instead it should be linked from documentation here.  At least in the ATSAMD case, I think the logical place for it is with the examples for what we call BSPs ([Metro M0 examples](https://github.com/atsamd-rs/atsamd/tree/master/boards/metro_m0/examples)).

[...]

PR #101, opened 2022-08-06
Add Development+Testing section to README.md

ianrrees:
See https://github.com/rust-embedded-community/usb-device/issues/100

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,15 @@ Class crates
 * [usbd-hid](https://github.com/twitchyliquid64/usbd-hid) [![Crates.io](https://img.shields.io/crates/v/usbd-hid.svg)](https://crates.io/crates/usbd-hid) - HID class
 * [usbd-serial](https://github.com/mvirkkunen/usbd-serial) [![Crates.io](https://img.shields.io/crates/v/usbd-serial.svg)](https://crates.io/crates/usbd-serial) - CDC-ACM serial port class
 
+Development
+-----------
+
+### Testing
+
+Testing `usb-device` involves a test suite running on a host computer, connected via USB to a target computer (microcontroller) which provides a test device.  `usb-device` provides interfaces to hardware drivers, but it does not provide any hardware drivers, so external crates are required to run the test suite on specific hardware:
+
+* [stm32-usbd-tests](https://github.com/Disasm/stm32-usbd-tests) for STM32 parts.
+* [test-usb-device](https://github.com/ianrrees/test-usb-device) for ATSAMD parts.
 
 Others
 ------


README.md
```diff
@@ -37,9 +37,18 @@ Class crates
...
 * [usbd-serial](https://github.com/mvirkkunen/usbd-serial) [![Crates.io](https://img.shields.io/crates/v/usbd-serial.svg)](https://crates.io/crates/usbd-serial) - CDC-ACM serial port class
 
+Development
+-----------
+
+### Testing
+
+Testing `usb-device` involves a test suite running on a host computer, connected via USB to a target computer (microcontroller) which provides a test device.  `usb-device` provides interfaces to hardware drivers, but it does not provide any hardware drivers, so external crates are required to run the test suite on specific hardware:
+
+* [stm32-usbd-tests](https://github.com/Disasm/stm32-usbd-tests) for STM32 parts.
```
Disasm, line 47:
```suggestion
* [stm32-usbd-tests](https://github.com/Disasm/stm32-usbd-tests) and [usb-otg-workspace](https://github.com/Disasm/usb-otg-workspace) for STM32 parts.
```

ianrrees: Ah, right, thanks!

README.md
```diff
@@ -37,9 +37,18 @@ Class crates
...
 
+Development
+-----------
+
+### Testing
+
+Testing `usb-device` involves a test suite running on a host computer, connected via USB to a target computer (microcontroller) which provides a test device.  `usb-device` provides interfaces to hardware drivers, but it does not provide any hardware drivers, so external crates are required to run the test suite on specific hardware:
+
+* [stm32-usbd-tests](https://github.com/Disasm/stm32-usbd-tests) and [usb-otg-workspace](https://github.com/Disasm/usb-otg-workspace) for STM32 parts.
+* [test-usb-device](https://github.com/ianrrees/test-usb-device) for ATSAMD parts.
```
ryan-summers, line 48: Should we instead just incorporate them as a member of the workspace here, so they don't get outdated?

ianrrees: Maybe, that's partly what I'm wondering about in #100 .

I'd imagine the overwhelming majority of users of usb-device aren't going to be interested in this device-specific test code.  As someone who spends more time tinkering with code that implements usb-device traits, rather than usb-device itself, I'd prefer to have the device-specific usb-device test code with whatever HAL it needs.  That HAL already will have usb-device as a dependency, which I believe covers the versioning concerns.

That said, I think we could bring together the documentation from these usb-device test repos, because that's mostly about testing usb-device in general.  I could imagine that documentation and test app expanding to cover other topics that might span multiple implementations of usb-device, for instance performance or feature sets.

ryan-summers: The `tests/` folder here isn't intended for users of usb-device, this is purely for CI and maintaince - as such, I don't think we should consider them here.

While having device-specific code isn't the purpose of `usb-device`, we ultimately still need to verify the functionality of `usb-device`, which at this moment requires hardware. As such, I think having HITL (hardware-in-the-loop) test code situated here is a big bonus. This will give developers a platform to bring their own targets up for testing if they so require.

I worry that pointing to separate repos for the device-specific test code will result in the situation that we **currently** have in the future where the test code is lost and no one knows how to test :P

ianrrees: > The tests/ folder here isn't intended for users of usb-device, this is purely for CI and maintaince - as such, I don't think we should consider user concerns here.

My comment could've been written better.  With "users". I was referring to both the people who would use `usb-device` (perhaps indirectly) in their projects, and people who are interested in changing/testing `usb-device` and code that interfaces with it.  Maybe something like "consumers" and "developers" could be used for these two groups - I think we should keep both in mind.

I agree that the details of how testing works probably don't matter to the consumers, but think it is still important to not make things worse for them.  Putting stuff in usb-device that won't be used by consumers is, to a small degree, making things worse for them - it's bloat that manifests in wasted bandwidth and disk.  Part of the question is how that presumably-small impact to many people compares to the impact to a small number of developers.


> I worry that pointing to separate repos for the device-specific test code will result in the situation that we currently have in the future where the test code is lost and no one knows how to test :P

The test code I made was a low-effort thing (in terms of `usb-device`) to support #60 - I knew basically how testing 'usb-device' worked at a technical level, my question about testing in #60 was more about expectations around testing proposed changes to 'usb-device'.  That said, I knew how the testing technically worked because I'd looked at the test source code - the obvious room for improvement there is about improving documentation for testing `usb-device`.

PR #103, opened 2022-08-09
Turn repository into workspace and separate the tests

eldruin:
Reading through #100 and #101, here is a PR turning this repository into a workspace. The idea would be that if @ianrrees and @Disasm agree, we could incorporate their device-side testing suite implementations into this repository so that everything is in a single place.
The directory structure would be:
```
|- usb-device                      // main crate
|- tests                           // test suite
    |- test_class_host             // host-side program
    |- <stm32xxxx>_device_tests    // device-side impl for some stm32 devices
    |- <atsamdxxxxx>_device_tests  // device-side impl for some atsamd devices
```
For now I just linked the repositories in the readme like in #101.
This includes the text from #101 as well. I hope that is fine with @ianrrees.
I do not care about the names of the device test folders. Feel free to propose a new one.
Thoughts?


ryan-summers (approved):
LGTM. This is what I had in mind as well, thanks for going through the work of restructuring! :)

eldruin:
Opinions @ianrrees and @Disasm?

ianrrees:
> This includes the text from https://github.com/rust-embedded-community/usb-device/pull/101 as well. I hope that is fine with @ianrrees.
Yep, totally fine!

I'll aim to put some more thoughts in #100 later today.

ianrrees:
Just a thought: what if we set up the host side test program as a normal bin?  I think that could allow a HAL developer to do something like `cargo install usb-device` to get the host-side test suite.

Disasm:
This looks like a better solution compared to what we have now, thank you!

Finomnis:
This is over a year old, what happened?

ryan-summers:
I believe the desire was lost and it now has a bunch of conflicts. usb-device isn't super well maintained - I review PRs in my spare time, but don't have do any active development on it.

ianrrees:
Exactly, @ryan-summers - Rust USB device stuff has been on my "next month or two" list for about two years now...  What we have is a good start, but it could definitely use some focused attention.

Finomnis:
> Rust USB device stuff has been on my "next month or two" list for about two years now...

Understandable, though. I'll also revert back to "good enough" once my usecase is fulfilled. :/

PR #104, opened 2022-08-19
Fix compiler warnings

haata:
- field xxx is never read
- associated funct xxx is never used
- Used _ to ignore the used fields and function


eldruin (approved):
Looks good to me, thanks!

PR #108, opened 2022-11-13
Fix link to embassy-usb

jannic:
The old link returned 404 as the file driver.rs was removed. Point to the crate root directory instead.

Alternatively, the link could point to the docs at https://docs.embassy.dev/embassy-usb/git/default/index.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,4 +42,4 @@ Others
 ------
 
 Other implementations for USB in Rust
-* [embassy-usb](https://github.com/embassy-rs/embassy/tree/master/embassy-usb) an async variant.
+* The [Embassy](https://github.com/embassy-rs/embassy) project has an async USB stack, embassy-usb.
--- a/README.md
+++ b/README.md
@@ -42,4 +42,4 @@ Others
 ------
 
 Other implementations for USB in Rust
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,4 +42,4 @@ Others
 ------
 
 Other implementations for USB in Rust
-* [embassy-usb](https://github.com/embassy-rs/embassy/blob/master/embassy-usb/src/driver.rs), an async variant.
\ No newline at end of file
+* [embassy-usb](https://github.com/embassy-rs/embassy/tree/master/embassy-usb) an async variant.
```
ryan-summers, line 45: Should we just point to top-level embassy instead? That way we don't break links if embassy restructures code

jannic: Good point. And I guess embassy-usb can not easily be used stand-alone in a project which is otherwise not using embassy (or can it?). So a sentence like "The [Embassy](https://embassy.dev/dev/index.html) project has an async USB stack." might also be more helpful to the reader. Should it point to the [docs](https://embassy.dev/dev/index.html) or to the [repository](https://github.com/embassy-rs/embassy)?

ryan-summers: I would just point to the repo and the user can go from there

ryan-summers (approved):
Thanks for the fix!

PR #116, opened 2023-04-19
Add the ability to select USB revision

ithinuel:
In some usecases it is desirable to select a lower revision of the USB standard (eg when a KVM between the device and the host does not handle usb 2.1 very well).

This allows a device to use a different usb revision (choices limited by an enum).

### DIFF
--- a/src/descriptor.rs
+++ b/src/descriptor.rs
@@ -110,12 +110,12 @@ impl DescriptorWriter<'_> {
         self.write(
             descriptor_type::DEVICE,
             &[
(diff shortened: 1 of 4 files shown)


Disasm:
If I recall correctly, BOS descriptor should be disabled for 2.0.

Disasm (approved):
Thank you!

ryan-summers:
@Disasm mind adding a changelog for this PR? I'm on mobile until next week, but can do it then if you can't get to it :)

PR #119, opened 2023-05-03
add reference to rp2040-hal and a few more class implementation crates.

ithinuel:


ryan-summers (approved):
Wonderful additions, thanks! :)

PR #120, opened 2023-06-23
Fix 404s in README

piotrfila:
Some of the links to class crates were outdated (404s). This PR makes them point to the repositories currently specified on crates.io.

(Side note: this is the second time I'm submitting this PR, I think something went wrong on my end. Sorry if this is a duplicate!)


ryan-summers (approved):
Thanks for catching this :)

PR #121, opened 2023-07-20
modify `StringIndex` and `.iad()`

eZioPan:
impl `u8` to `StringIndex`, remove `StringIndex::new()`, pop `StringIndex` for `.iad()`

such that user can define interface/iad name easily.

### DIFF
--- a/src/descriptor.rs
+++ b/src/descriptor.rs
@@ -185,10 +185,13 @@ impl DescriptorWriter<'_> {
         function_class: u8,
         function_sub_class: u8,
         function_protocol: u8,
+        function_string: Option<StringIndex>,
     ) -> Result<()> {
         if !self.write_iads {
             return Ok(());
         }
+
+        let str_index = function_string.map_or(0, Into::into);
 
         self.write(
             descriptor_type::IAD,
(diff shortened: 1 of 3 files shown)


src/descriptor.rs
```diff
@@ -185,11 +185,14 @@ impl DescriptorWriter<'_> {
         function_class: u8,
         function_sub_class: u8,
         function_protocol: u8,
+        function_string: Option<StringIndex>,
```
ryan-summers, line 188: Can you add docs for this as well?

ryan-summers (changes requested):
Thanks for the change!

Can you add a [breaking] entry into CHANGELOG.md?

mvirkkunen:
The idea of the private constructor was to make sure string indexes are uniquely allocated via the allocator. Did that make something hard/impossible to implement?

eZioPan:
> The idea of the private constructor was to make sure string indexes are uniquely allocated via the allocator. Did that make something hard/impossible to implement?

got it, i will reverse the change for `StringIndex`, thus user will use `alloc.string()` to query a new index.

src/descriptor.rs
eZioPan: done

ryan-summers (approved):
Thanks, looking good. Thanks for the input @mvirkkunen :)

eZioPan:
thanks!

PR #122, opened 2023-07-21
multi-language STRING Desc support

eZioPan:
add a ~~`extra_lang_ids: Option<&'a [u16]>`~~ `extra_lang_ids: Option<&'a [LangID]>` to `struct Config` to store LANGIDs.

extend `descriptor_type::STRING` branch of `fn get_descriptor()`, so that it will send all support LANGID in first STRING Desc, and try responding every STRING Request with matched LANGID.

add a ~~`.extra_lang_ids()`~~ `.set_extra_lang_ids()` method to `UsbDeviceBuilder`, which will take a ~~`&[u16]`~~ `&[LangID]` to add extra support LANGID.

**Breaking**

`UsbDeviceBuilder::manufacturer()` `UsbDeviceBuilder::product()` `UsbDeviceBuilder::serial_number()` will ask a `&[&str]` as input, first `&str` in `&[&str]` will match English lang, the rest will match what have been put into ~~`.extra_lang_ids()`~~ `.set_extra_lang_ids()`.


src/device.rs
```diff
@@ -63,9 +63,10 @@ pub(crate) struct Config<'a> {
     pub product_id: u16,
     pub usb_rev: UsbRev,
     pub device_release: u16,
-    pub manufacturer: Option<&'a str>,
-    pub product: Option<&'a str>,
-    pub serial_number: Option<&'a str>,
+    pub extra_lang_ids: Option<&'a [u16]>,
+    pub manufacturer: Option<&'a [&'a str]>,
+    pub product: Option<&'a [&'a str]>,
+    pub serial_number: Option<&'a [&'a str]>,
```
ryan-summers, line 69: I'm assuming the order of entries in the lists corresponds to the order in `extra_lang_ids`. If so, we need to specify that somehow to the user.

src/device.rs
```diff
@@ -547,25 +548,73 @@ impl<B: UsbBus> UsbDevice<'_, B> {
...
-                            let index = StringIndex::new(index);
-                            let lang_id = req.index;
+            descriptor_type::STRING => match index {
+                // first STRING Request
+                0 => {
+                    if let Some(extra_lang_ids) = config.extra_lang_ids {
+                        assert!(
+                            extra_lang_ids.len() < 16,
+                            "Not support more than 15 extra LangIDs"
+                        );
```
ryan-summers, line 558: This should be done when adding lang IDs to the config, not at run-time. Please move this assert elsewhere.

src/device.rs
```diff
@@ -547,25 +548,73 @@ impl<B: UsbBus> UsbDevice<'_, B> {
...
+                        let mut lang_id_bytes = [0u8; 32];
+                        lang_id_bytes[0..2].copy_from_slice(&lang_id::ENGLISH_US.to_le_bytes());
+
+                        extra_lang_ids
+                            .iter()
+                            .enumerate()
+                            .for_each(|(index, lang_id)| {
+                                lang_id_bytes[(index + 1) * 2..(index + 2) * 2]
+                                    .copy_from_slice(&lang_id.to_le_bytes());
+                            });
```
ryan-summers, line 569: This is much simpler by using rust's fancy iterator logic. You may need to clean this up a bit to make it valid rust, but it should look as follows:

```suggestion
                        let mut lang_id_bytes = [0u8; 32];

                        for (buffer, lang_id) in lang_id_bytes.chunks_exact_mut(2).zip(&[lang_id::ENGLISH_US].iter().chain(extra_lang_ids.iter())) {
                            buffer.copy_from_slice(&lang_id.to_le_bytes());
                        }
```

src/device_builder.rs
```diff
@@ -106,27 +107,60 @@ impl<'a, B: UsbBus> UsbDeviceBuilder<'a, B> {
         self
     }
 
+    /// Sets **extra** Language ID for device.
+    ///
+    /// Since "en_US"(0x0409) is implicitly embedded, you just need to fill other LangIDs
+    ///
+    /// Default: (none)
+    pub fn extra_lang_ids(mut self, extra_lang_ids: &'a [u16]) -> Self {
```
ryan-summers, line 115: Can we name this something more ergonomic, like `add_extra_languages()` or similar?

src/device_builder.rs
```diff
@@ -106,27 +107,60 @@ impl<'a, B: UsbBus> UsbDeviceBuilder<'a, B> {
...
+        assert!(
+            extra_lang_ids.len() < 16,
+            "Not support more than 15 extra LangIDs"
+        );
+
+        if extra_lang_ids.len() > 0 {
+            self.config.extra_lang_ids = Some(extra_lang_ids);
+        }
+
+        self
```
ryan-summers, line 125: After this update, check `manufacturer`, `product`, and `serial_number` to ensure they have the correct number of slice entries. The number of extra-lang-ids and the number of elements in these slices must always be equal.

Obviously only do this if those fields are Some (can't check before they're set)

src/device_builder.rs
```diff
@@ -106,27 +107,60 @@ impl<'a, B: UsbBus> UsbDeviceBuilder<'a, B> {
...
+    /// the first string should always be in English, the language of reset strings
+    /// should be pair with what inside [.extra_lang_ids()](Self::extra_lang_ids)
+    ///
     /// Default: (none)
-    pub fn manufacturer(mut self, manufacturer: &'a str) -> Self {
-        self.config.manufacturer = Some(manufacturer);
+    pub fn manufacturer(mut self, manufacturer_ls: &'a [&'a str]) -> Self {
+        if manufacturer_ls.len() > 0 {
+            self.config.manufacturer = Some(manufacturer_ls);
+        }
```
ryan-summers, line 137: Check that the number of lang strings provided is equal to extra-lang-ids + 1, throw an error otherwise.

src/device_builder.rs
```diff
@@ -106,27 +107,60 @@ impl<'a, B: UsbBus> UsbDeviceBuilder<'a, B> {
...
     /// Sets the product name string descriptor.
     ///
+    /// the first string should always be in English, the language of reset strings
+    /// should be pair with what inside [.extra_lang_ids()](Self::extra_lang_ids)
+    ///
     /// Default: (none)
-    pub fn product(mut self, product: &'a str) -> Self {
-        self.config.product = Some(product);
+    pub fn product(mut self, product_ls: &'a [&'a str]) -> Self {
+        if product_ls.len() > 0 {
```
ryan-summers, line 148: Check that the number of lang strings provided is equal to extra-lang-ids + 1, throw an error otherwise.

src/device_builder.rs
```diff
@@ -106,27 +107,60 @@ impl<'a, B: UsbBus> UsbDeviceBuilder<'a, B> {
...
     /// Sets the serial number string descriptor.
     ///
+    /// the first string should always be in English, the language of reset strings
+    /// should be pair with what inside [.extra_lang_ids()](Self::extra_lang_ids)
+    ///
     /// Default: (none)
-    pub fn serial_number(mut self, serial_number: &'a str) -> Self {
-        self.config.serial_number = Some(serial_number);
+    pub fn serial_number(mut self, serial_number_ls: &'a [&'a str]) -> Self {
+        if serial_number_ls.len() > 0 {
```
ryan-summers, line 161: Check that the number of lang strings provided is equal to extra-lang-ids + 1, throw an error otherwise.

ryan-summers (changes requested):
Looking really cool. Glad to see we're adding support for non-English language :)

Some minor feedback requested to make this more robust

src/device.rs
eZioPan: I have doc the relationship of 4 lists in 3 methods of the `UsbDeviceBuilder`. Should I also put a comment on `Config` struct too?

ryan-summers: I saw the documentation after reviewing this, I think what we have already is appropriate

src/device.rs
eZioPan: I will remove this `assert!`, I think the `assert!` in the methods of `UsbDeviceBuilder` will be enough.

src/device_builder.rs
ryan-summers, line 115: Also, it would be incredibly useful if we just enumerated the `LangId` as an enum that was exposed to the user, so they didn't have to go explicitly look up all the lang-id values

eZioPan: Since this function override the extra language id list, I might think `set_extra_languages()` a better one?

src/device_builder.rs
eZioPan: If we check lang id count in this function, we assume user set `extra_lang_id` before `manufacturer`, `product`, and `serial_number`. I'm not sure whether we should force/assume user will do this or not.

ryan-summers: No, we just don't check if `extra_lang_id` is None. We'll check later when they actually specify the extra lang IDs

src/device_builder.rs
eZioPan: ok, I will try dump the LANGID list from [MS-LCID] into the enum after I modify other parts.

[MS-LCID]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-lcid/70feba9f-294e-491e-b6eb-56532684c37f

ryan-summers: If it starts looking like a big task, don't worry about it and spawn an issue to make it for later :)

eZioPan: got it, thanks.

src/device.rs
eZioPan: done

src/device_builder.rs
eZioPan: done

eZioPan:
hi, I have do some modifications.
And `LangID` implementation is really ugly, but I know little about macro_rule, would you give me some hints?

src/device_builder.rs
eZioPan: done

src/device_builder.rs
eZioPan: done

src/device_builder.rs
eZioPan: done

src/descriptor/lang_id.rs
```diff
@@ -0,0 +1,940 @@
...
+    }
+}
+
+impl From<LangID> for u16 {
+    fn from(lang_id: LangID) -> Self {
+        lang_id as u16
+    }
+}
+
+impl TryFrom<u16> for LangID {
```
ryan-summers, line 15: This duplicates all of the LangIDs and makes it easy to accidentally break things. Why not use a readily-available tool like [`num_enum::TryFromPritive`](https://docs.rs/num_enum/latest/num_enum/derive.TryFromPrimitive.html)?

src/device.rs
```diff
@@ -547,25 +549,69 @@ impl<B: UsbBus> UsbDevice<'_, B> {
...
+                            3 => config
+                                .serial_number
+                                .map(|str_list| str_list[lang_id_list_index]),
+
+                            _ => unreachable!(),
                         }
+                    } else {
+                        // for other custom STRINGs
+                        let index = StringIndex::new(index);
+                        let lang_id = req.index.try_into().unwrap();
```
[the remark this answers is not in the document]
ryan-summers: Instead, this should be propogated outwards and/or logged and/or rejected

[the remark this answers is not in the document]
eZioPan: I will try reject the request and write down a log(don't know how to propograte outward).
about the log, will `defmt::warn!()` sufficient?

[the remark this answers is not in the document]
ryan-summers: Yeah, any log would be fine if there's existing logging infrastructure in the crate. If not, just reject the transfer.

src/descriptor/lang_id.rs
eZioPan: done

src/device.rs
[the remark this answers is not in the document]
eZioPan: done

src/device_builder.rs
```diff
@@ -106,27 +108,132 @@ impl<'a, B: UsbBus> UsbDeviceBuilder<'a, B> {
...
+            "Not support more than 16 \"manufacturer\"s"
+        );
+
+        // do list length check only if user already specify "extra_lang_ids"
+        if let Some(extra_lang_ids) = self.config.extra_lang_ids {
+            assert!(
+                manufacturer_ls.len() == extra_lang_ids.len() + 1,
+                "The length of \"manufacturer\" list should be one more than \"extra_lang_ids\" list",
+            )
+        }
```
ryan-summers, line 173: This check needs to be done even if there are no extra lang IDs. I.e.
```suggestion
        let num_extra_langs = self.config.extra_lang_ids.as_ref().map(|langs| langs.len()).unwrap_or(0);
        
        assert!(
            manufacturer_ls.len() == num_extra_langs + 1,
            "The length of \"manufacturer\" list should be one more than \"extra_lang_ids\" list",
        )

```

src/device_builder.rs
```diff
@@ -106,27 +108,132 @@ impl<'a, B: UsbBus> UsbDeviceBuilder<'a, B> {
...
+            return self;
+        }
+
+        assert!(
+            product_ls.len() <= 16,
+            "Not support more than 16 \"product\"s"
+        );
+
+        // do list length check only if user already specify "extra_lang_ids"
+        if let Some(extra_lang_ids) = self.config.extra_lang_ids {
```
ryan-summers, line 198: This check needs to be done regardless of whether or not extra_lang_ids is Some

src/device_builder.rs
```diff
@@ -106,27 +108,132 @@ impl<'a, B: UsbBus> UsbDeviceBuilder<'a, B> {
...
+            return self;
+        }
+
+        assert!(
+            serial_number_ls.len() <= 16,
+            "Not support more than 16 \"serial_number\"s"
+        );
+
+        // do list length check only if user already specify "extra_lang_ids"
+        if let Some(extra_lang_ids) = self.config.extra_lang_ids {
```
ryan-summers, line 228: This check needs to be done regardless of whether or not extra_lang_ids is Some

ryan-summers (changes requested):
Looking really good - one last request

eZioPan:
hi, I still got a few concerns.

1. Currently the max count of lang_id is limit to **16**, which isn't force by USB Spec. I pick this number "randomly". I'm not sure if it's a good idea.
   * If this limitation is ok, then maybe we need to specify this number as a `const` value, such that we can easily modify it when some one meets the limit. This number has been used in `device_buider.rs` and `device.rs`(to specify `lang_id_bytes` length), so maybe write a `const` in `lib.rs` file? What would you sugguest?
2. The variants of `LangID` enum is dumped from [MS-LCID]. I did some modification, I'm not very sure.
   1. I manually remove every LANGID mark as **Reserved**, **Neither defined nor reserved**, **temporarily assigned**, etc.
       I keep the LANGID both marked with a language and **Reserved**.
   2. I blindly replace hyphen with underscore, and make every character upper case(since `0x004D` named `as`, happens to be a keyword of rust).
   3. `ff-NG` and `ff-Latn-NG` occupy the same LANGID `0x0467`, I concat them to `FF_NG__FF_LATN_NG`
3. Different Host USB stack may have different LANGID implemtations, and since LANGID has been marked as "deprecated" both by [USB.org] and [MS]. Should we doc something on `LangID` enum? 

[MS-LCID]: https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-lcid/70feba9f-294e-491e-b6eb-56532684c37f
[USB.org]: https://usb.org/deprecated-links-and-tools
[MS]: https://learn.microsoft.com/en-us/windows/win32/intl/language-identifier-constants-and-strings

src/device_builder.rs
eZioPan: done

ithinuel (commented):
IMHO, it'd be simpler/less costly to check that `EN_US` is part of the custom `lang_ids` list at the time it is set, rather than having constructs like `[LangID::EN_US].iter().chain(extra_lang_ids.iter())`.

See: https://github.com/eZioPan/usb-device/compare/multi-language-support...ithinuel:usb-device:PR122

src/device.rs
```diff
@@ -547,24 +549,99 @@ impl<B: UsbBus> UsbDevice<'_, B> {
...
+                                            req_lang_id
+                                        );
+                                        None
+                                    })
+                                    .and_then(|lang_id_list_index| {
+                                        match index {
+                                            1 => config.manufacturer,
+                                            2 => config.product,
+                                            3 => config.serial_number,
+                                            _ => unreachable!(),
```
ithinuel, line 633: `index == 0` is not unreachable.

eZioPan:
Hi @ithinuel , thanks for the work. I think it's a good solution. Let see what @ryan-summers will say about this.

src/device_builder.rs
eZioPan: I re-think about this. And I think this code may force user call `.set_extra_lang_ids()` before other methods.

This code will panic

```rust
builder.product(&["name1", "name2"]).set_extra_ids(&[lang_id1])
```

If we need this, we may doc the "calling" order of the methods.

ithinuel: Or use a different builder type, e.g. with a generic marker.

ithinuel:
A modified my proposal to check `LangID::EN_US` is first.
I also added a commit that should lift the arbitrary limit to the number of languages.

eZioPan:
@ithinuel it looks really cool. I have test on my device, and find that there could be a minor bug.
In `src/device.rs` Line 560, currently it is `if len < buf.len()`, I think it should be `if buf.len() < len` ?

src/device_builder.rs
ryan-summers: Ah you're right @eZioPan. In general, I think the builder needs to change like @ithinuel notes, but I'd prefer if we did that in a different PR. Let's at least get _something_ in here :)

src/device_builder.rs
```diff
@@ -106,27 +108,144 @@ impl<'a, B: UsbBus> UsbDeviceBuilder<'a, B> {
...
-        self.config.serial_number = Some(serial_number);
+    pub fn serial_number(mut self, serial_number_ls: &'a [&'a str]) -> Self {
+        if serial_number_ls.len() == 0 {
+            self.config.serial_number = None;
+            return self;
+        }
+
+        assert!(
+            serial_number_ls.len() <= 16,
+            "Not support more than 16 \"serial_number\"s"
```
ryan-summers, line 232:
```suggestion
            "The length of the \""serial_number\" list should be one more than \"extra_lang_ids\" list
```

eZioPan:
I have reverse the change, keep 16 lang_ids limitation.

ryan-summers:
> Different Host USB stack may have different LANGID implemtations, and since LANGID has been marked as "deprecated" both by [USB.org](https://usb.org/deprecated-links-and-tools) and [MS](https://learn.microsoft.com/en-us/windows/win32/intl/language-identifier-constants-and-strings). Should we doc something on LangID enum?

Please do - if this is the case, I'm less concerned about the overall "correctness" of the various lang IDs - if people hit issues, they can submit PRs to fix it, I don't see a huge risk in a slightly incorrect lang ID outside of potentially failing to enumerate properly on some PC dialects, but given that lang ID is deprecated, I imagine this will go away in the future anyways.

ryan-summers (approved):

ryan-summers:
If you want to add documentation to LangID, feel free. If not, I'll merge as-is in a little while

eZioPan:
Thanks, I would ask a merge. I'm afraid that I'm not very good at English, so maybe I just leave it.
What I can provide is that currently this LANGID is dumped from **MS-LCID Protocol Revision 15.0**, so if anyone find a bug, their can fetch some information from the MS side.

PR #126, opened 2023-08-02
Update CHANGELOG.md

### LINKED ISSUE #125: Changelog: add another `EndpointType` changelog to "Breaking" part
https://github.com/rust-embedded-community/usb-device/blob/7ebb10fe5f952955b416c5b05abc84d37bf82acf/CHANGELOG.md?plain=1#L22-L23

Since `EndpointType` is a public enum, and some crates [may use the integer form of it](https://github.com/stm32-rs/synopsys-usb-otg/blob/aa12f1a4e21441159a1f3a646768f43cbb3e74b9/src/endpoint.rs#L92-L99). After this modification user will need to call `.to_bm_attributes()` to access numeric form. We might add a another changelog to tell user about this.

```diff
### Breaking
+++ * Acess numeric form of `EndpointType` variants now require a `.to_bm_attributes()` ([#60](https://github.com/rust-embedded-community/usb-device/pull/60)).
```

eZioPan:
will close #125

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
   * `UsbDeviceBuilder` has a public `.extra_lang_ids()` method to specify LANGIDs besides ENGLISH_US(0x0409)
 
 ### Breaking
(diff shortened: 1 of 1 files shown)


eldruin (approved):
Thank you!

PR #127, opened 2023-09-11
Add copy and eq to UsbError

Finomnis:
`UsbError` is as primitive as it gets, and there are situations where the error has to be duplicated (for example to put it in an error event in a non-blocking queue).

Unless there are grand plans for the `UsbError` enum that will add references to data in the future, it would help me to add `Clone`. The other ones are mostly suggestions based on best practices.


ryan-summers (commented):
Looks good to me, thanks for the change. Mind adding a CHANGELOG entry?

Finomnis:
@ryan-summers Done.

ryan-summers (approved):

PR #130, opened 2023-11-02
Refactoring language management for simplicity

ryan-summers:
This PR builds on #122 by unifying strings under a single `StringDescriptors` structure to keep things organized a little differently. This eases usage on the end-user, as previously the borrowed slices had to be `&'static` borrows, which complicated the construction when i.e. a serial number was read from EEPROM or some other source.

### DIFF
--- a/src/device_builder.rs
+++ b/src/device_builder.rs
@@ -23,6 +23,59 @@ macro_rules! builder_fields {
     }
 }
 
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum BuilderError {
+    TooManyLanguages,
+    InvalidPacketSize,
+    PowerTooHigh,
+}
+
+/// Provides basic string descriptors about the device, including the manufacturer, product name,
+/// and serial number of the device in a specified language.
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub struct StringDescriptors<'a> {
+    pub(crate) id: LangID,
+    pub(crate) serial: Option<&'a str>,
+    pub(crate) product: Option<&'a str>,
+    pub(crate) manufacturer: Option<&'a str>,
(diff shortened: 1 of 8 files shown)


ryan-summers:
Additionally, I'd like to release after this lands - this was an issue I encountered while testing the new API for a 0.3 release

src/device_builder.rs
```diff
@@ -108,133 +158,17 @@ impl<'a, B: UsbBus> UsbDeviceBuilder<'a, B> {
...
-            )
-        }
-
-        self.config.serial_number = Some(serial_number_ls);
-
-        self
+    /// # Note
+    /// Up to 16 languages may be provided.
+    pub fn strings(mut self, descriptors: &[StringDescriptors<'a>]) -> Result<Self, BuilderError> {
+        // The 16 language limit comes from the size of the buffer used to provide the list of
```
ianrrees, line 166: Super-minor nit/question: does the size of that buffer ([this one](https://github.com/rust-embedded-community/usb-device/pull/130/files#diff-36327182a60f061fcbee1b4da122b99e82b6d4d213e4b3e94bf297354680a720R550)?) have documentation that explains this consequence?  Does the limit originate in the USB spec?

ithinuel (approved):

src/device_builder.rs
ryan-summers: I actually don't think language IDs are even _in_ the specification anymore, so there's not really anything limiting more than 16 languages. It's from USB 1.0 from what I can tell and has largely been deprecated (not sure in favor of what). Even finding the list of lang IDs is hard nowadays. It makes me wonder if we should even support multiple langs, but I guess it makes sense for legacy systems?

PR #133, opened 2023-11-13
Provide `LangID` via preludes

ianrrees:
It seems like most uses of usb-device now require the LangID enum, so this PR adds it to the preludes.


ryan-summers (approved):
Thanks, was thinking this myself

PR #134, opened 2023-11-13
Documentation updates

ianrrees:
If https://github.com/rust-embedded-community/usb-device/pull/133 is accepted, this should have the explicit `use` for `LangID` removed from the example code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,14 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+## [0.3.1] - 2023-11-15
+
 ### Added
 * `BuilderError`, `LangID`, `StringDescriptors` now in `prelude`
 * `LangID` now in `class_prelude`
+
+### Changed
+* Updated documentation, including example code
 
 ## [0.3.0] - 2023-11-13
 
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -149,7 +149,9 @@ pub mod endpoint;
 /// // product name. If using an existing class, remember to check the class crate documentation
 /// // for correct values.
 /// let mut usb_dev = UsbDeviceBuilder::new(&usb_bus, UsbVidPid(0x5824, 0x27dd))
-///     .product(&["Serial port"])
+///     .strings(&[StringDescriptors::new(LangID::EN)
+///         .product("Serial port")
+///         .expect("Failed to set strings")
```
ryan-summers, line 154:
```suggestion
///     .strings(&[StringDescriptors::new(LangID::EN)
///         .product("Serial port")
///         .expect("Failed to set strings")
///     ])
```

ryan-summers: Yeah, it should

ianrrees: Thanks!  I've force-pushed a fix.

ianrrees:
I've force-pushed changes to bump the patch version.

Also can report that the tests pass on ATSAMD21 and ATSAMD51 boards when using a branch of the HAL that has a couple minor updates to work with the newer usb-device.  Once the HAL changes are merged, I'll aim to get back to https://github.com/rust-embedded-community/usb-device/issues/100

ithinuel:
Hello, can we use this opportunity to make BuilderError public or does it need it's own PR?

ianrrees:
Sounds good to me!  @ithinuel does the new commit work for your purpose?

CHANGELOG.md
```diff
@@ -5,10 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
```
Disasm, line 8: It might be a good idea to leave an empty `Unreleased` section in the changelog

ithinuel:
It does, thank you 👍

ryan-summers (approved):
Lgtm. I'll release here when I get in the office :) thanks for the updates!

ianrrees:
Sorry to throw this in at the last minute (and end of a long day for me - apologies if this doesn't make sense!) but I'm wondering if `BuilderError` is maybe too generic of a name, for something that is put in the prelude - lots of things use Builder...  Maybe `UsbBuilderError`, or a super-verbose `UsbDeviceBuilderError`?

ryan-summers:
If people don't like the name, they can rename it during import. The same argument can be made for `Error`

PR #136, opened 2023-11-28
Update dependecies

Sh3Rm4n:
This fixes an issue, with `proc-macro2` in a minimal version build: `cargo +nightly update -Z minimal-versions`

Found out at https://github.com/stm32-rs/stm32f3xx-hal/actions/runs/7016287690/job/19087185399?pr=349, which pulls proc-macro `v1.0.55`

(this crate might however not be the cause of this issue)

`num_enum_derive` has a newer version, which fixes the issue https://github.com/illicitonion/num_enum/blob/a80fa1ef9d1c23995650e9446b1b0721289a2cbd/num_enum_derive/Cargo.toml#L35


ryan-summers (approved):
LGTM, Thanks for the changes!

PR #137, opened 2023-12-15
Add `free` function to UsbAllocator

jnthbdn:
Allowing users to free the UsbBus from the UsbAllocator.

### DIFF
--- a/src/bus.rs
+++ b/src/bus.rs
@@ -316,6 +316,11 @@ impl<B: UsbBus> UsbBusAllocator<B> {
         self.alloc(None, EndpointType::Interrupt, max_packet_size, interval)
             .expect("alloc_ep failed")
     }
+
+    /// Free the UsbBus
+    pub fn free(self) -> B {
+        self.bus.into_inner()
+    }
 }
 
 /// A handle for a USB interface that contains its number.


eldruin (commented):
I am not sure it makes sense to allow this.
Maybe somebody else can comment like @ryan-summers ?

ryan-summers:
Yeah I'm also not sure this makes sense to allow. Can you elaborate on what the necessary use case is here?

jnthbdn:
Hello,

Thank you for your feedback. I've added this function to make the link with the PR725 (https://github.com/rp-rs/rp-hal/pull/725) from rp-hal.
The problem is that creating a `UsbBus` requires ownership of two peripherals (`USBCTRL_REGS` and `USBCTRL_DPRAM`) and a clock. But if you want to stop the USB, it's not possible to recover these resources, and you end up, for example, with a `Partially move` error (in my case from the clockManager).

ianrrees:
It would be nice to have the ability to tear down structures - I could imagine this being useful for implementing a low power device that usually keeps the USB peripheral powered down, for instance.

However, I see two issues with this implementation:

1. Any allocated endpoints will essentially be left dangling.  So, I think there would need to be destructors added to the endpoints, and some bookkeeping so that the free methods behave correctly.
2. `UsbDevice::build()` calls `UsbBusAllocator::freeze()`, so it seems like the new `free()` wouldn't work as desired after `build()`.

PR #140, opened 2024-01-31
Make LangID optional on get_string()

jose-acevedoflores:
I've been playing around with the [Microsoft OS String Descriptors](https://github.com/pbatard/libwdi/wiki/WCID-Devices#microsoft-os-string-descriptor) (specifically to get the winusb driver installed on my device) and everything was working fine on 0.2.9 but the update to 0.3.0 (#122) introduced a behavior difference. 

The change in #122 requires that  `req.index` be a valid lang id, if not, it exits early and doesn't call the `get_string` method. 

I also thought lang id could be set to a default in order to not get a breaking change but I think it obscures that LangId might not be set for all requests? 

Any thoughts appreciated. I'm just dabbling on embedded stuff so I could be missing something.

### DIFF
--- a/src/class.rs
+++ b/src/class.rs
@@ -44,7 +44,7 @@ pub trait UsbClass<B: UsbBus> {
     /// * `index` - A string index allocated earlier with
     ///   [`UsbAllocator`](crate::bus::UsbBusAllocator).
     /// * `lang_id` - The language ID for the string to retrieve.
-    fn get_string(&self, index: StringIndex, lang_id: LangID) -> Option<&str> {
+    fn get_string(&self, index: StringIndex, lang_id: Option<LangID>) -> Option<&str> {
         let _ = (index, lang_id);
         None
     }
--- a/src/class.rs
+++ b/src/class.rs
@@ -43,8 +43,9 @@ pub trait UsbClass<B: UsbBus> {
     ///
     /// * `index` - A string index allocated earlier with
     ///   [`UsbAllocator`](crate::bus::UsbBusAllocator).
-    /// * `lang_id` - The language ID for the string to retrieve.
-    fn get_string(&self, index: StringIndex, lang_id: Option<LangID>) -> Option<&str> {
+    /// * `lang_id` - The language ID for the string to retrieve. If the requested lang_id is not
+    ///   valid it will default to EN_US.
+    fn get_string(&self, index: StringIndex, lang_id: LangID) -> Option<&str> {
         let _ = (index, lang_id);
         None
     }
--- a/src/device.rs
+++ b/src/device.rs
@@ -563,25 +563,21 @@ impl<B: UsbBus> UsbDevice<'_, B> {
 
                 // rest STRING Requests
                 _ => {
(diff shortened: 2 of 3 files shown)


src/device.rs
```diff
@@ -563,22 +563,25 @@ impl<B: UsbBus> UsbDevice<'_, B> {
...
+                                    #[cfg(feature = "defmt")]
+                                    defmt::warn!(
+                                        "Receive unknown LANGID {:#06X}, reject the request",
+                                        _err.number
+                                    );
+                                    xfer.reject().ok();
+                                    return;
+                                }
+
+                                Ok(req_lang_id) => req_lang_id,
```
ryan-summers, line 582: Does this code not introduce the same problem that you're describing in the PR? I.e. if the lang_id is invalid for the string IDs 1..3, wouldn't this also reject the transfer unexpectedly?

src/device.rs
```diff
@@ -599,7 +602,7 @@ impl<B: UsbBus> UsbDevice<'_, B> {
                             let index = StringIndex::new(index);
                             classes
                                 .iter()
-                                .find_map(|cls| cls.get_string(index, lang_id))
+                                .find_map(|cls| cls.get_string(index, lang_id.ok()))
```
ryan-summers, line 605: I think defaulting to EN_US for the lang ID is probably a good idea. I don't think Lang ID is even defined in the USB spec anymore.

src/class.rs
```diff
@@ -44,7 +44,7 @@ pub trait UsbClass<B: UsbBus> {
     /// * `index` - A string index allocated earlier with
     ///   [`UsbAllocator`](crate::bus::UsbBusAllocator).
     /// * `lang_id` - The language ID for the string to retrieve.
-    fn get_string(&self, index: StringIndex, lang_id: LangID) -> Option<&str> {
+    fn get_string(&self, index: StringIndex, lang_id: Option<LangID>) -> Option<&str> {
```
ryan-summers, line 47: I'm very hesitant about introducing another API change in the traits. It results in a ton of downstream work for various crates to re-release and is still not done for the 0.3.0 version.

src/device.rs
jose-acevedoflores: you are correct, it does reject for IDs 1, 2 and 3. I just limited the changes of #122 to those 3 requests since I thought  they were "standard" and the lang_id would be expected to be valid.

src/class.rs
jose-acevedoflores: yeah I get it, I can go the simpler route and default to EN_US if that is ok.

src/device.rs
ryan-summers: I think that'd be best

src/class.rs
ryan-summers: Lets go with that for now.

jose-acevedoflores:
Made the change to just use `EN_US` as fallback if LangID is not valid. Let me know if you want me to collapse the two commits into one too.

ryan-summers:
Wonderful, thanks for the PR! :)

PR #141, opened 2024-02-18
Convert LangID from u16 enum to tuple struct containing u16

### LINKED ISSUE #139: Improvement: Minimize LANGID enum under feature flag
Recently I started mitigating my code to use usb-device 0.3.1 on an stm32l4 project.

looking the objdump I noticed the langid conversion method requires ~8KB of code:
image
clearly, mapping a huge enum this is unavoidable.
however,
to support smaller builds I think hiding most langIDs under a feature flag would suffice.

as a test run I copied the langid enum and covered it with a feature flag `min-lang-id`

```rust
#[allow(missing_docs)]
#[derive(Clone, Copy, PartialEq, Debug, TryFromPrimitive)]
#[cfg_attr(feature = "defmt", derive(defmt::Format))]
#[repr(u16)]
#[cfg(feature="min-lang-id")]
pub enum LangID {
    EN_US = 0x0409,
}
```

assuming we minimize to only EN_US, the enum gets completly optimized out.
I'm sure there can be a minimal set of common LANGIDs actually seen in the wild.
Regardless, I do think that avoiding adding this much code for an embedded targeted code is essential.

vitalyvb:
Relax lang_id checks provided by enum so that any lang_id could be used, not only values listed in the enum. These checks increased the code size considerably without much benefit. Resolves #139.

Language identifiers like LangID::EN_US have a new type, but this should not require source code changes in a code which uses them.

Provide two traits to convert LangID to and from u16 freely. This implicitly provides TryFrom trait which previously was derived using TryFromPrimitive.

There's no special constructor for LangID for values outside of the provided constants, but it can be converted from any u16.

As TryFromPrimitive in no longer used, remove num_enum dependency.


ryan-summers (approved):
Thanks for this PR, I was just starting to look into this and forgot someone had submitted a PR for it already.

While this isn't _technically_ a semver-compatible change without bumping the minor version (since some of the `LangID` API has changed), I don't think anyone was actively using that API since it was rather new. Given that, I'm fine just releasing this as a patch version and we can yank it later if it causes anyone trouble.

We need a CHANGELOG entry here, but I don't mind adding it myself later. Thanks for doing the leg work here. I'm going to test it out on some hardware and merge if it works :)

ryan-summers:
It worked great for me and definitely reduced the code size by 8KB upon analysis. Thanks for the change!

PR #142, opened 2024-02-26
Adding USB tracing, fixing further enumeration issues

ryan-summers:
This PR does a few things:
1. Adds in a new logging feature to make it easy to trace USB debug issues in the future (I've written this a few times, and figured it would be useful as a general feature instead)
2. Removed the `set_error()` calls when processing a SETUP packet. We are not allowed to error in this state and should simply ignore the SETUP packet instead.
3. Changed the handling of EP0-IN-complete tokens. We were being pretty jumpy with getting these tokens in unexpected states, however this is simply an indication that a packet transmission completed. We may have updated state internally and lost track of what we sent. As such, I've updated it such that we just ignore these tokens if we don't need them in our current state instead of stalling the control pipe, as this isn't really an error.

This effectively reverts https://github.com/rust-embedded-community/usb-device/pull/41, as further updates to EP0 control pipe handling make it unnecessary. Reversion of this PR prevents us from sending more data in a control transfer than is expected, since the host may have initiated the STATUS phase of the transfer already.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+* A new `log` feature can be enabled to provide logging and tracing information about the USB
+interface.
+
 ### Changed
 * Invalid LangIDs will default to `EN_US`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ interface.
 
 ### Changed
 * Invalid LangIDs will default to `EN_US`
+* Changed handling of EP0 state to eliminate unexpected issues with device enumeration
 
 ## [0.3.1] - 2023-11-15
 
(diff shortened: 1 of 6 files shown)


ryan-summers:
@eldruin / @thejpster / @mvirkkunen I've tested this with 2 STM devices (Booster and Stabilizer, an STM32H7 and an STM32F4) successfully, and it reportedly fixed one of our users issues where the device was not enumerating properly (ref https://github.com/quartiq/stabilizer/issues/817), so I think it'd be beneficial to get this landed and re-released as a potential 0.3.1 :)

jannic:
I successfully updated one of the rp2040 examples (https://github.com/rp-rs/rp-hal-boards/blob/773ee73bfdff66d080dda528bfd03db845b75967/boards/rp-pico/examples/pico_usb_twitchy_mouse.rs) to use this version of usb-device. That probably doesn't test many code paths, and I only have linux hosts to test it with, but at least this makes sure that your changes don't break USB on the rp2040 completely.

eldruin (approved):
Looks fine to me but I do not know how it works :)
You can decide whether to wait for an opinion from @mvirkkunen.

ryan-summers:
I'm going to merge this for now and aim for releasing soon to give some feedback period from anyone wanting to test it out. Thanks for taking a look everyone, I know USB can be extremely opaque - I'm no expert myself.

PR #143, opened 2024-02-28
Handling `Result`s throughout the crate

### LINKED ISSUE #58: Errors during get_configuration_descriptors get silently eaten
https://github.com/mvirkkunen/usb-device/blob/258ad5b9e44b6ef9b0b7fa55254b12db0a53fc3f/src/device.rs#L413-L417

If `get_configuration_descriptors` throws an error (e.g. because of a buffer overflow), the error will exit the `.accept()` call and then get dropped by the `.ok()` (which will convert errors to `None` and drop the error). This makes it somewhat more difficult to debug when the configuration descriptors are too long (for example).

I'm not entirely sure what the correct behaviour should be. One option would be to `unwrap()` and panic (which is probably reasonable here, since it'll always panic essentially when the device boots, so it's easy to detect). The other option would be to make `get_descriptor` (and therefore `control_in`) return `Result`, which is probably the much more invasive solution (it'd be difficult to integrate with various class drivers).

### LINKED ISSUE #113: Replace usage of `expect()` with returning `Result<>` types
Currently, there's a number of places where `expect()` is called on `Result<>` types. This can introduce format bloat into end-user applications, which may be undesirable.

Instead, we should propagate the `Result<>` out to the user to allow them to decide what to do with the error.

ryan-summers:
Fixes #113 by handling all the various `Result<>` types and propogating them upwards to `poll()`. For now, they are simply logged. In the future, it could be helpful to push these outwards and have `poll()` return the `Result<>` type, however I'm hesitant to introduce a breaking API change right now. I'll spawn an issue for this.

This also fixes #58 by propagating the errors described to the new log macros.

### DIFF
--- a/src/control_pipe.rs
+++ b/src/control_pipe.rs
@@ -136,10 +136,11 @@ impl<B: UsbBus> ControlPipe<'_, B> {
                 let count = match self.ep_out.read(&mut self.buf[i..]) {
                     Ok(count) => count,
                     Err(UsbError::WouldBlock) => return Ok(None),
-                    Err(_) => {
+                    Err(_err) => {
                         // Failed to read or buffer overflow (overflow is only possible if the host
                         // sends more data than it indicated in the SETUP request)
                         self.set_error();
+                        usb_debug!("Failed EP0 read: {:?}", _err);
                         return Ok(None);
                     }
                 };
--- a/src/control_pipe.rs
+++ b/src/control_pipe.rs
@@ -129,18 +129,18 @@ impl<B: UsbBus> ControlPipe<'_, B> {
         None
     }
 
-    pub fn handle_out(&mut self) -> Option<Request> {
+    pub fn handle_out(&mut self) -> Result<Option<Request>> {
         match self.state {
             ControlState::DataOut(req) => {
                 let i = self.i;
                 let count = match self.ep_out.read(&mut self.buf[i..]) {
                     Ok(count) => count,
-                    Err(UsbError::WouldBlock) => return None,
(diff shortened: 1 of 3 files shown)


eldruin (approved):
Looks good to me, thank you!
I guess this does not need an entry in the changelog if the changes are all internal.

mvirkkunen:
Just out of interest, did you make sure that e.g. the state changes in `ControlPipe` still work correctly even though the functions may return early now? I seem to remember that there were already some problems in there because error handling was "too strict" and in a lot of cases it might have been better to just ignore some of them because implementations can be racey and/or broken.

ryan-summers:
> Just out of interest, did you make sure that e.g. the state changes in `ControlPipe` still work correctly even though the functions may return early now? I seem to remember that there were already some problems in there because error handling was "too strict" and in a lot of cases it might have been better to just ignore some of them because implementations can be racey and/or broken.

It's hard to say for sure or not because most of the errors would originate in the actual PHY implementation. That being said, I've tested this on both Linux and Windows with an STM32F4 and an H7, and both enumerated properly with both Release and Debug builds.

I think most of the racey and/or broken communication issues were control-pipe related and got fixed up in https://github.com/rust-embedded-community/usb-device/pull/142

While there's a possibility that issues could be getting introduced, the new logging features should make them simpler to work through when they occur.

ryan-summers:
Going to merge this in favor of moving forwards, but will definitely keep an eye out for regressions, thanks for bringing this up @mvirkkunen.

PR #144, opened 2024-02-28
Constructing the control pipe from a user-provided buffer

### LINKED ISSUE #31: Make the control transfer buffer size actually configurable
Currently there is only a hack to double its size. Ideally the user should be able to supply their own `&'static [u8]` to use for control transfers - however this is easier said than done.

Design goals:

1. Enable the user to supply their own buffer...
2. ...but still somehow have a default buffer that is used automatically if the user doesn't supply one.
3. Don't waste memory by having the default buffer allocated even when not used.

I tried to deal with "3" by using a generic argument for the control buffer type. It threaded through a lot of types and was too ugly to justify (having a generic argument on `UsbDevice` just for this is excessively verbose)

ryan-summers:
This PR updates the `UsbDevice` to use a buffer for the control pipe that is provided by the user. This fixes #31.

This is a breaking API change so would require an eventual 0.4.0 release, which would be a good time to additionally propagate errors outwards etc. in the API.

### DIFF
--- a/src/test_class.rs
+++ b/src/test_class.rs
@@ -22,6 +22,8 @@ mod sizes {
     pub const INTERRUPT_ENDPOINT: u16 = 31;
 }
 
+static mut CONTROL_BUFFER: [u8; 256] = [0; 256];
+
 /// Test USB class for testing USB driver implementations. Supports various endpoint types and
 /// requests for testing USB peripheral drivers on actual hardware.
 pub struct TestClass<'a, B: UsbBus> {
@@ -112,7 +114,7 @@ impl<B: UsbBus> TestClass<'_, B> {
         &self,
         usb_bus: &'a UsbBusAllocator<B>,
     ) -> UsbDeviceBuilder<'a, B> {
-        UsbDeviceBuilder::new(usb_bus, UsbVidPid(VID, PID))
+        UsbDeviceBuilder::new(usb_bus, UsbVidPid(VID, PID), unsafe { &mut CONTROL_BUFFER })
             .strings(&[StringDescriptors::default()
                 .manufacturer(MANUFACTURER)
                 .product(PRODUCT)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ interface.
 as this was not intended to be used in the public API.
     - If this is problematic, please open an issue in the main `usb-device` repository.
 * Changed handling of EP0 state to eliminate unexpected issues with device enumeration
+* [breaking] The control pipe is now provided in the `UsbDeviceBuilder` API to allow for user-provided control
+pipes. This makes it so that control pipes have configurable sizing.
 
 ## [0.3.1] - 2023-11-15
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,8 +20,6 @@ rusb = "0.9.1"
 rand = "0.8.5"
 
 [features]
-# Use a 256 byte buffer for control transfers instead of 128.
-control-buffer-256 = []
 
 # Enable logging and tracing via the `log` crate
 log = ["dep:log"]
--- a/src/control_pipe.rs
+++ b/src/control_pipe.rs
@@ -20,31 +20,28 @@ enum ControlState {
     Error,
 }
 
-// Maximum length of control transfer data stage in bytes. 128 bytes by default. You can define the
-// feature "control-buffer-256" to make it 256 bytes if you have larger control transfers.
-#[cfg(not(feature = "control-buffer-256"))]
-const CONTROL_BUF_LEN: usize = 128;
-#[cfg(feature = "control-buffer-256")]
-const CONTROL_BUF_LEN: usize = 256;
-
 /// Buffers and parses USB control transfers.
 pub struct ControlPipe<'a, B: UsbBus> {
     ep_out: EndpointOut<'a, B>,
     ep_in: EndpointIn<'a, B>,
     state: ControlState,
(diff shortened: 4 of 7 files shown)


eldruin (commented):
Thank you! Could you have a look at the CI?

ryan-summers:
Fixed CI and reverted this to a draft - I'd like to release before we make breaking API changes here.

vitalyvb:
Perhaps it would be nice to have a check that the buffer is at least as large as max_packet_size_0.
Something like
```diff
diff --git a/src/device_builder.rs b/src/device_builder.rs
index c32c03d..d582737 100644
--- a/src/device_builder.rs
+++ b/src/device_builder.rs
@@ -33,6 +33,8 @@ pub enum BuilderError {
     InvalidPacketSize,
     /// Configuration specifies higher USB power draw than allowed
     PowerTooHigh,
+    /// Control buffer is smaller than max_packet_size_0
+    SmallControlBuffer,
 }
 
 /// Provides basic string descriptors about the device, including the manufacturer, product name,
@@ -110,8 +112,12 @@ impl<'a, B: UsbBus> UsbDeviceBuilder<'a, B> {
     }
 
     /// Creates the [`UsbDevice`] instance with the configuration in this builder.
-    pub fn build(self) -> UsbDevice<'a, B> {
-        UsbDevice::build(self.alloc, self.config, self.control_buffer)
+    pub fn build(self) -> Result<UsbDevice<'a, B>, BuilderError> {
+        if self.control_buffer.len() < self.config.max_packet_size_0 as usize {
+            Err(BuilderError::SmallControlBuffer)
+        } else {
+            Ok(UsbDevice::build(self.alloc, self.config, self.control_buffer))
+        }
     }
 
     builder_fields! {
diff --git a/src/test_class.rs b/src/test_class.rs
index 70dbb9c..e08736a 100644
--- a/src/test_class.rs
+++ b/src/test_class.rs
@@ -96,7 +96,7 @@ impl<B: UsbBus> TestClass<'_, B> {
 
     /// Convenience method to create a UsbDevice that is configured correctly for TestClass.
     pub fn make_device<'a>(&self, usb_bus: &'a UsbBusAllocator<B>) -> UsbDevice<'a, B> {
-        self.make_device_builder(usb_bus).build()
+        self.make_device_builder(usb_bus).build().unwrap()
     }
 
     /// Convenience method to create a UsbDeviceBuilder that is configured correctly for TestClass.
```

Or just a
```diff
diff --git a/src/device_builder.rs b/src/device_builder.rs
index c32c03d..02320e3 100644
--- a/src/device_builder.rs
+++ b/src/device_builder.rs
@@ -111,6 +111,9 @@ impl<'a, B: UsbBus> UsbDeviceBuilder<'a, B> {
 
     /// Creates the [`UsbDevice`] instance with the configuration in this builder.
     pub fn build(self) -> UsbDevice<'a, B> {
+        if self.control_buffer.len() < self.config.max_packet_size_0 as usize {
+            panic!("control_buffer is smaller than max_packet_size_0");
+        }
         UsbDevice::build(self.alloc, self.config, self.control_buffer)
     }
 
```

src/test_class.rs
```diff
@@ -112,7 +114,7 @@ impl<B: UsbBus> TestClass<'_, B> {
         &self,
         usb_bus: &'a UsbBusAllocator<B>,
     ) -> UsbDeviceBuilder<'a, B> {
-        UsbDeviceBuilder::new(usb_bus, UsbVidPid(VID, PID))
+        UsbDeviceBuilder::new(usb_bus, UsbVidPid(VID, PID), unsafe { &mut CONTROL_BUFFER })
```
vitalyvb, line 117: Not an issue really, but I get the following warning with 1.77.0-nightly
```
warning: mutable reference of mutable static is discouraged
   --> .../usb-device/src/test_class.rs:117:70
    |
117 |         UsbDeviceBuilder::new(usb_bus, UsbVidPid(VID, PID), unsafe { &mut CONTROL_BUFFER })
    |                                                                      ^^^^^^^^^^^^^^^^^^^ mutable reference of mutable static
    |
    = note: for more information, see issue #114447 <https://github.com/rust-lang/rust/issues/114447>
    = note: reference of mutable static is a hard error from 2024 edition
    = note: mutable statics can be written to by multiple threads: aliasing violations or data races will cause undefined behavior
    = note: `#[warn(static_mut_ref)]` on by default
help: mutable references are dangerous since if there's any other pointer or reference used for that static while the reference lives, that's UB; use `addr_of_mut!` instead to create a raw pointer
    |
117 |         UsbDeviceBuilder::new(usb_bus, UsbVidPid(VID, PID), unsafe { addr_of_mut!(CONTROL_BUFFER) })
    |                                                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

ryan-summers:
@vitalyvb Great idea on checking the control buffer against the max packet size. I've added these checks to the builder.

src/test_class.rs
ryan-summers: Hmm, this is indeed a strange problem to solve for embedded. Using the mentioned macro would force us to change the API to accept a raw pointer instead of a slice reference, which really tanks usability for a non-`static mut` buffer. Let me take a look at this and figure out the right approach.

ryan-summers: Based on the comments on the referenced issue, the approach was to use an `UnsafeCell` around the buffer, which is very doable. I have updated the implementation and docs. Thanks for pointing this out :)

vitalyvb (approved):
Looks good. Thanks!

PR #145, opened 2024-02-28
Restore deriving defmt::Format for LangID when defmt feature is enabled

vitalyvb:
defmt deriving for LangID was unintentionally removed by pull request #141. Restore it.


ryan-summers (approved):
Awesome, thanks :)

Issue #32, opened 2020-02-03
UsbClass::poll not always called on UsbDevice::poll


TimNN:
Contrary to what is documented on `UsbClass::poll` (*"Called whenever the `UsbDevice` is polled."*), it is only called by `UsbDevice::poll` when `UsbBus::poll` returns `PollResult::Data`.

Is that working as intended with the documentation being wrong or does the documentation need to be updated?

TimNN:
Maybe to provide a bit of context: I've been working USB Serial Logger implementation. The logs are initially written to an internal buffer. Whenever `endpoint_in_complete` is called, the buffer is checked and if logs are present they are written.

The issue is, of course, when there are no logs to be written. In that case `endpoint_in_complete` will never be called again. Based on the documentation, I implemented `poll` so that it would periodically also check the buffer and trigger an initial write if necessary. I few hours of debugging later, I noticed that my `poll` method wasn't actually called 😑.

Disasm:
Slightly offtopic: have you tried `usbd-serial` for this task?

TimNN:
The Logger is actually a wrapper around `usbd-serial`.

Disasm:
Then it should start sending when you `write` to it.

TimNN:
> Then it should start sending when you write to it.

It does. The issue is that the initial write is never triggered because it happens during `UsbClass::poll`.

Disasm:
@mvirkkunen, any thoughts? I think that the current behavior is correct and the documentation is wrong, but who knows.

mvirkkunen:
Hmm.. good question! Calling it only when data is received is somewhat redundant because the specific methods (`control_in`, `endpoint_out`) etc will also have been called at that point if the class needs to know about data.

*However* the correct way to implement this would be to proactively write into the endpoint outside of `poll`, even if `UsbClass::poll` was called on every poll. The reason is that it's OK for users to only call  `UsbDevice::poll` in a USB interrupt handler, which may not be executed at all if there is no traffic. `usbd-serial` writes a packet immediately after the first buffer is written into the port this and while this may result in the first write of a batch being inefficient (possibly only one byte), other implementations would require either help from the user (a guaranteed periodic call to a method) or some kind of generic timer system which is perhaps out of scope.

Honestly I don't remember what `UsbClass::poll` was meant to be used for. It might be a candidate for removal now to avoid confusion.

TimNN:
[...]

PR #147, opened 2024-03-11
Updating poll functions to return results

ryan-summers:
This PR updates the traits to allow `poll()` to return `Result<>` types if needed. I'm unsure if the `UsbClass::poll()` function should return a `Result` or not currently.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -172,13 +172,13 @@ impl<B: UsbBus> UsbDevice<'_, B> {
     ///
     /// Strictly speaking the list of classes is allowed to change between polls if the device has
     /// been reset, which is indicated by `state` being equal to [`UsbDeviceState::Default`].
-    pub fn poll(&mut self, classes: &mut ClassList<'_, B>) -> bool {
+    pub fn poll(&mut self, classes: &mut ClassList<'_, B>) -> Result<bool> {
         let pr = self.bus.poll();
 
         if self.device_state == UsbDeviceState::Suspend {
             match pr {
                 PollResult::Suspend | PollResult::None => {
-                    return false;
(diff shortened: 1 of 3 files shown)


sourcebox:
I think this PR makes sense because currently there's no way to return an error from the driver. Using `PollResult::None` in such situations is ambiguous.

src/device.rs
```diff
@@ -228,15 +228,17 @@ impl<B: UsbBus> UsbDevice<'_, B> {
 
                     match req {
                         Some(req) if req.direction == UsbDirection::In => {
-                            if let Err(_err) = self.control_in(classes, req) {
+                            if let Err(err) = self.control_in(classes, req) {
                                 // TODO: Propagate error out of `poll()`
```
ianrrees, line 232: This comment should go away with the PR, the same appears in a couple more places below.

src/device.rs
```diff
@@ -317,10 +319,10 @@ impl<B: UsbBus> UsbDevice<'_, B> {
                 }
 
                 for cls in classes.iter_mut() {
-                    cls.poll();
```
ianrrees, line 320: This introduces a subtle change in behaviour that maybe should be documented: if there are multiple classes and an earlier one errors out, the later classes won't be polled.

However, looking at the current implementation and docstring, it's not explicit that polling the device (sometimes... unclear if that's correct per https://github.com/rust-embedded-community/usb-device/issues/32) results in the classes being polled.

ianrrees (approved):
I like this change, in particular because it encourages users to think about how to handle the return value of poll() - in most of the examples I've seen it is just ignored.

PR #148, opened 2024-04-13
Feature gate portable atomics

MarcusGrass:
Allow an opt-out for portable atomics, if the platform has atomic load/store then they aren't necessary

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ repository = "https://github.com/rust-embedded-community/usb-device"
 
 [dependencies]
 defmt = { version = "0.3", optional = true }
-portable-atomic = { version = "1.2.0", default-features = false }
+portable-atomic = { version = "1.2.0", default-features = false, optional = true }
 heapless = "0.8"
 log = { version = "0.4", default-features = false, optional = true}
 
@@ -20,6 +20,7 @@ rusb = "0.9.1"
 rand = "0.8.5"
 
 [features]
+default = ["portable-atomic"]
 
 # Use larger endpoint buffers for highspeed operation (default fullspeed)
 #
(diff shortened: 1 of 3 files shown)


mvirkkunen:
Doesn't `portable-atomic` already do feature detection and do essentially the same thing automatically?

MarcusGrass:
> Doesn't `portable-atomic` already do feature detection and do essentially the same thing automatically?
> 
> Edit: Okay, I guess it can be a bit overzealous sometimes if the target doesn't support _all_ atomic operations, but it has its own feature flags to disable that it seems. Could that be used instead?

I think that from a codegen standpoint it'll just be the same in release, and the unused stuff **should** all be eliminated, but from a hygiene perspective fewer dependencies would be nicer because of the usual reasons, not really that important though. 

I fully wanted this out of my dependency graph to make sure that there weren't any unintended side-effects like intrinsics creation that I relied on implicitly and decided might as well post a patch. 
Although, since i noticed a major is coming up, making it opt-in would be even better (from my perspective). But it's just a matter of taste.

ryan-summers:
While I understand the desire to have `portable-atomics` out of your individual dependency graph, I'm not sure this is the best approach for the broader audience. The whole point of the `portable-atomics` dependency is to make atomics work for the user regardless of platform and to enable this crate for all platforms as painlessly as possible.

Given that the end result from a codegen perspective _should_ be the same for platforms where atomics are supported, I don't see a benefit in forcing users to figure out whether or not their platform has atomics and selecting the appropriate feature flags.

I fully understand that wanting to ensure emulated atomics aren't present is a very real design task for many projects, I don't think that this is the best approach for the wide-ranging audience that this crate supports.

PR #151, opened 2024-05-27
Add usbd-midi to class crates list in readme

sourcebox:
As suggested on Matrix, this PR adds usbd-midi to the list of class crates in the readme file.


eldruin (approved):
Sounds good to me, thanks!

PR #153, opened 2024-08-10
Add new read_setup() for control endpoints

ianrrees:
Background can be found at https://github.com/atsamd-rs/atsamd/pull/738 .

The USB spec says that the SETUP transactions used with control endpoints must always be accepted by the device [1], and at least in the Microchip SAMD21's USB peripheral that behaviour is handled in hardware [2].  We've observed that some hosts will send an OUT ZLP, followed very shortly by a SETUP for a new transaction - on the SAMD21 parts this results in the OUT data (none, in this case) being overwritten by SETUP, which in the existing implementation results in an overflow error (since `read()` was called with a zero length slice for the OUT ZLP, but >0 bytes are available from the SETUP that overwrote it).  The user-observable result is that the device fails to enumerate at all on some hosts, or sporadically on others.

In light of this, it seems the control pipe code needs some way to differentiate reads of OUT data from reads of SETUP data.  This PR changes the existing `read()` methods to return an error if SETUP data is present, making it specific to OUT data, and adds a new `read_setup()` method specifically for reading SETUP data.

Most of the commits/diff here are work on the test framework, to add a new test that resets the device and times how long it takes to re-enumerate.  The test is a bit crude, but seems to validate the changes to the read methods.

[1] USB 2.0 spec section 5.5.5 Control Transfer Data Sequences:
> If a Setup transaction is received by an endpoint before a previously initiated control transfer is completed,
> the device must abort the current transfer/operation and handle the new control Setup transaction. A Setup
> transaction should not normally be sent before the completion of a previous control transfer. However, if a
> transfer is aborted, for example, due to errors on the bus, the host can send the next Setup transaction
> prematurely from the endpoint’s perspective.

[2] SAMD21 datasheet DS40001882H section 32.6.2.6 Management of SETUP Transactions.  The key thing here is that the data is copied in to the buffer regardless of BK0RDY.

### DIFF
--- a/src/endpoint.rs
+++ b/src/endpoint.rs
@@ -1,5 +1,5 @@
 use crate::bus::UsbBus;
-use crate::{Result, UsbDirection};
+use crate::{Result, UsbDirection, UsbError};
 use core::marker::PhantomData;
 use portable_atomic::{AtomicPtr, Ordering};
 
@@ -197,9 +197,10 @@ impl<B: UsbBus> Endpoint<'_, B, In> {
 }
 
 impl<B: UsbBus> Endpoint<'_, B, Out> {
-    /// Reads a single packet of data from the specified endpoint and returns the actual length of
-    /// the packet. The buffer should be large enough to fit at least as many bytes as the
-    /// `max_packet_size` specified when allocating the endpoint.
+    /// Reads a single packet of data and returns the length of the packet
+    ///
+    /// The buffer should be large enough to fit at least as many bytes as the `max_packet_size`
+    /// specified when allocating the endpoint.
     ///
     /// # Errors
     ///
@@ -211,8 +212,36 @@ impl<B: UsbBus> Endpoint<'_, B, Out> {
     ///   USB. A zero-length packet will return `Ok(0)`.
     /// * [`BufferOverflow`](crate::UsbError::BufferOverflow) - The received packet is too long to
     ///   fit in `data`. This is generally an error in the class implementation.
+    /// * [`InvalidState`](crate::UsbError::InvalidState) - The received packet is a SETUP
+    ///   transaction, and needs to be read through [`read_setup()`](Endpoint::read_setup())
+    ///   instead.
     pub fn read(&self, data: &mut [u8]) -> Result<usize> {
         self.bus().read(self.address, data)
+    }
+
+    /// Reads a single packet of SETUP data and returns the length of the packet
+    ///
+    /// The buffer should be large enough to fit at least as many bytes as the `max_packet_size`
+    /// specified when allocating the endpoint.  See [`UsbBus::read_setup()`] for rationale for two
+    /// distinct read methods.
+    ///
+    /// # Errors
+    ///
+    /// Note: USB bus implementation errors are directly passed through, so be prepared to handle
+    /// other errors as well.
+    ///
+    /// * [`WouldBlock`](crate::UsbError::WouldBlock) - There is no packet to be read. Note that
+    ///   this is different from a received zero-length packet, which is valid and significant in
+    ///   USB. A zero-length packet will return `Ok(0)`.
+    /// * [`BufferOverflow`](crate::UsbError::BufferOverflow) - The received packet is too long to
+    ///   fit in `data`. This is generally an error in the class implementation.
+    pub fn read_setup(&self, data: &mut [u8]) -> Result<usize> {
+        // SETUP transactions can only occur on control endpoints
+        if self.ep_type != EndpointType::Control {
+            Err(UsbError::InvalidEndpoint)
+        } else {
+            self.bus().read_setup(self.address, data)
+        }
     }
 }
 
--- a/src/endpoint.rs
+++ b/src/endpoint.rs
@@ -199,49 +199,56 @@ impl<B: UsbBus> Endpoint<'_, B, In> {
 impl<B: UsbBus> Endpoint<'_, B, Out> {
     /// Reads a single packet of data and returns the length of the packet
     ///
-    /// The buffer should be large enough to fit at least as many bytes as the `max_packet_size`
-    /// specified when allocating the endpoint.
+    /// The buffer should be large enough to fit at least as many bytes as the
+    /// `max_packet_size` specified when allocating the endpoint.
     ///
     /// # Errors
     ///
-    /// Note: USB bus implementation errors are directly passed through, so be prepared to handle
-    /// other errors as well.
+    /// Note: USB bus implementation errors are directly passed through, so be
+    /// prepared to handle other errors as well.
     ///
-    /// * [`WouldBlock`](crate::UsbError::WouldBlock) - There is no packet to be read. Note that
-    ///   this is different from a received zero-length packet, which is valid and significant in
-    ///   USB. A zero-length packet will return `Ok(0)`.
-    /// * [`BufferOverflow`](crate::UsbError::BufferOverflow) - The received packet is too long to
-    ///   fit in `data`. This is generally an error in the class implementation.
-    /// * [`InvalidState`](crate::UsbError::InvalidState) - The received packet is a SETUP
-    ///   transaction, and needs to be read through [`read_setup()`](Endpoint::read_setup())
-    ///   instead.
+    /// * [`WouldBlock`](crate::UsbError::WouldBlock) - There is no OUT packet
+    ///   to be read. Note that this is different from a received zero-length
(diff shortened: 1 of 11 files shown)


ianrrees:
@ithinuel - this is the PR discussed the other day in Matrix, I've put the corresponding atsamd-rs changes [here](https://github.com/ianrrees/atsamd/tree/enumeration), and test class implementations for SAMD21 and SAMD51 are [here](https://github.com/ianrrees/test-usb-device/tree/enumeration)

ianrrees:
@9names ran some tests of this with Raspberry Pi 5 host and SAMD21, confirming the problem (enumeration being unreliable) and that this appears to resolve it.

While confirming the problem, 9names saw the new test case error out in an unexpected way - reporting `Entity not found` - I think that might require a bit more work to make the enumeration time test more reliable.

ianrrees:
Sorry, I've not had an opportunity to look at the test code changes here, and probably won't be able to test Rust USB stuff until mid December at earliest.  Since the test stuff isn't really critical, I'd like to proceed with this PR so that implementors can do their part.

rnd-ash:
Out of interest, what is blocking this PR? As it is required to get USB functioning on the ATSAMD platform properly.

ianrrees:
@eldruin @ryan-summers - Apologies for my sporadic involvement with this, I'd like to get it over the line but it's not clear to me what the next step should be.  The important change is about the API change, but in order to demonstrate the problem I made some fairly large changes to the testing infrastructure, maybe those should be separated?

Also, I'm hesitant to offer as I've not been doing much programming lately, but I could try to help out with maintaining usb-device by reviewing other PRs/issues if that might help share the load a bit.  Would prefer to not accept my own PRs of course!

eldruin (commented):
Thank you for your work on this.
1. Have you looked deeper into the `Entity not found` test error? I hesitate to merge a test that may fail for unrelated and unclear reasons when another PR is submitted.
2. IIUC, this is a breaking change, so we would need to release a 0.4 version, which will lead to a lot of friction across the ecosystem given the position of this crate. Do you have an impression of the breakage and solutions when attempting to mix `usb-device` versions in applications and HALs? 
3. One thing missing here is an entry to the changelog.
4. I would welcome you to the org if you are interested to join. Mostly we do 4-eye principle so that at least one more person has a look at the changes. This is not a formal policy, though. Have a look [here](https://github.com/rust-embedded-community) for further info.

ianrrees:
1. No, sorry I've not followed up.  From memory, enumeration failures+retries are the most obvious symptom of the problem that the API change aims to fix, so the new test triggers hard resets of the device and measures the time that the OS takes to enumerate the device.  I don't believe that libusb provides an API that would allow a more direct test of the device behaviour.  The test case that failed is new and quite different type of test from what was here before, which is a part of why I'm thinking of separating the test changes (which might need more work) from the API change (which is relatively straightforward).
2. Yes, this is a difficult issue - a device needs to use exactly one version of usb-device, which means that all the USB class implementations, HALs, etc need to update when there's a new release usb-device.
3. Done, cheers!
4. Thanks!  If I can make more time for this sort of thing, I'll put my hand up to join as a maintainer.

ianrrees:
One thing I'm curious about: have other UsbBus implementations than the Microchip SAM HAL got some clever solution for this, with the existing usb-device?  My hunch is that it isn't possible to make a correctly-behaving device without some way to distinguish between reads of SETUP and OUT data, but my usb-device experience is limited to SAMD devices.

ithinuel:
I had some troubles with zlp followed by setup while implementing it on rp2040.
The device keeps flags for things it received. I'm on my phone rn but will edit when I get a chance to get back to my computer to link the impl we have.

EDIT:
We capture in the poll method if the setup flag was set [here](https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/src/usb.rs#L729-L738) in `inner.read_setup` and remember it until a read is issued on ep_0 [here](https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/src/usb.rs#L371-L372).

And if a ZLP got discarded by the hardware/software, we ignore it [here](https://github.com/rp-rs/rp-hal/blob/main/rp2040-hal/src/usb.rs#L394-L395).

ianrrees:
Interesting, thanks @ithinuel !  That is a clever workaround, and I think the need for it shows that the usb-device API has a bit of a gap in this area.  I'll rework this PR a little so it doesn't talk about the SAMD target, because your example shows the same underlying issue.

Presuming that the RP2040 USB peripheral works like the SAMD one, I'd note that the workaround leaves a possibly-large timing hole - if poll() is called after the ZLP, poll() returns, SETUP appears, then ep_read() is called.

ithinuel:
True, I realize that as I was looking at the code & thinking of your use case:

```
- -> ZLP
- poll()
- -> SETUP
- read(0, 0)
- poll()
- read(0, 8) <- would probably crash.
```

That being said, I never experienced any issue with it so far. I’m using the rp2040 as my keyboard through a KVM causing the usb to re-enumerate everytime I switch computer.
My guess would be that because the rp2040 runs at least 3 times faster than the SAMD21, so the poll-read sequence is superfast.

ianrrees:
When that happened in the SAMD case, the read() would fail (silently, IIRC) and the host would retry the SETUP unless it had failed too many times.  With most hosts, it was reliable enough to just be an annoyance (enumeration would take a second or two longer) but some never worked.

ianrrees:
Just a nudge on this - maybe good to integrate a couple outstanding PRs (eg https://github.com/rust-embedded-community/usb-device/pull/147) and cut a release.

My feeling is that, although it's a pain to update all the other crates, this is as the documentation states a work-in-progress and this PR allows the HAL implementations to fix incorrect behaviour.

src/endpoint.rs
```diff
@@ -211,9 +212,37 @@ impl<B: UsbBus> Endpoint<'_, B, Out> {
...
+    ///   instead.
     pub fn read(&self, data: &mut [u8]) -> Result<usize> {
         self.bus().read(self.address, data)
     }
+
+    /// Reads a single packet of SETUP data and returns the length of the packet
+    ///
+    /// The buffer should be large enough to fit at least as many bytes as the `max_packet_size`
+    /// specified when allocating the endpoint.  See [`UsbBus::read_setup()`] for rationale for two
+    /// distinct read methods.
```
kevinmehall: A SETUP packet is a fixed size of 8 bytes, so max_packet_size isn't relevant, and you could consider taking `&mut [u8; 8]` or even returning `Result<[u8; 8]>`.

src/endpoint.rs
```diff
@@ -211,9 +212,37 @@ impl<B: UsbBus> Endpoint<'_, B, Out> {
     ///   USB. A zero-length packet will return `Ok(0)`.
     /// * [`BufferOverflow`](crate::UsbError::BufferOverflow) - The received packet is too long to
     ///   fit in `data`. This is generally an error in the class implementation.
+    /// * [`InvalidState`](crate::UsbError::InvalidState) - The received packet is a SETUP
+    ///   transaction, and needs to be read through [`read_setup()`](Endpoint::read_setup())
+    ///   instead.
```
kevinmehall: Rather than treating this as a separate error case, would it make more sense to return `WouldBlock`? Just treat a `SETUP` as a totally different thing than an `OUT` because they're handled in separate places.

If I'm understanding the architecture of this library correctly, you already have the `ep_setup` flag triggering  `handle_setup` and resetting the state when receiving a SETUP packet. And that needs to happen regardless of whether `read` in `handle_out` has seen this error or not, so it's best if that happens without `handle_out` doing anything first. For example, it looks like in the `ControlState::DataOut` state, an error from `read` in `handle_out` calls `set_stall`, which means that if the host polls between then and when `handle_setup` resets it, you'd reject the new request.

And if you do have a reason to want that error to break out of something blocking on a control transfer completion or something, you'd probably want it on `write` as well.

kevinmehall (commented):
Overall this seems like the right direction.

src/endpoint.rs
ianrrees: Good point!

src/endpoint.rs
ianrrees: I think `WouldBlock` does make sense, and your understanding is correct AFAICT

ianrrees:
Spending a little more time with this today after a(nother) period of absence, I think that the test changes should be split out and worked on separately.  Have implemented changes suggested by @kevinmehall (thanks so much!) and the tests pass on a Metro M0 Express (SAMD21) but I'll aim to split this PR up in to two

PR #162, opened 2025-06-10
fix: impl `Sync` for `UsbBusAllocator`

bobozaur:
Fixes an issue that I think arose in https://github.com/rust-embedded-community/usb-device/pull/149. Because `UsbBus` no longer requires `Sync`, then the trait is no longer auto implemented.


bobozaur:
Nvm it seems that the `UsbBusAllocator` was `!Sync` even before #149 . Is that on purpose?

PR #167, opened 2025-09-11
bump defmt to v1, bump rand dev-dep to v0.9

robamu:


AudaciousAxiom:
I'm not a maintainer of this project, but this PR might be more straightforward to review if it were only about the bump of `defmt` to v1, as the bump of `rand` might require more consideration (even if it is only a dev-dependency). (Also the latest release of `rand` is v0.10 now).

Bumping `defmt` to v1 shouldn't be a breaking change as v0.3.100 implements the [SemVer trick](https://github.com/dtolnay/semver-trick): v0.3.100's re-exports the `defmt::Format` trait from v1, so the trait being used right now [is already that of v1](https://docs.rs/crate/defmt/0.3.100/source/src/lib.rs#330).

Additionally, the [announcement post of v1](https://ferrous-systems.com/blog/defmt-1-0/#:~:text=Any%20library,version,-%2E) explicitly says that it is safe to do so:

> Any library currently using defmt = "^0.3" in their Cargo.toml file can replace that line with defmt = "1", and everything will continue to work, unless you are locked to a version of 0.3 prior to the final 1.0-compatible version.

Cargo.toml
```diff
@@ -10,14 +10,14 @@ authors = ["Matti Virkkunen <mvirkkunen@gmail.com>"]
...
-defmt = { version = "0.3", optional = true }
+defmt = { version = "1", optional = true }
 portable-atomic = { version = "1.2.0", default-features = false }
 heapless = "0.8"
 log = { version = "0.4", default-features = false, optional = true}
 
 [dev-dependencies]
 rusb = "0.9.1"
-rand = "0.8.5"
+rand = "0.9"
```
eldruin:
```suggestion
rand = "0.10"
```

eldruin:
Done in #172 . Thank you for your work!

PR #168, opened 2025-09-11
allow heapless v0.9

robamu:
I think 0.8 also works, so this allows both versions.


eldruin (commented):
Sounds good to me, could you also add an entry to the changelog?

robamu:
done

eldruin (approved):
Thank you!

